PR #1, opened 2016-04-06
[JENKINS-34008] Add a hook listener for the global library

abayer:
[JENKINS-34008](https://issues.jenkins-ci.org/browse/JENKINS-34008)

Yup, it's back but this time against the new repo. =)

Existing rebuild test validates this change.

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryListener.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryListener.java
@@ -16,25 +16,16 @@
  */
 package org.jenkinsci.plugins.workflow.cps.global;
 
-import hudson.Extension;
 import hudson.ExtensionPoint;
-import javax.inject.Inject;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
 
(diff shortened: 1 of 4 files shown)


svanoort:
> Yup, it's back but this time against the new repo. =)

The global library can run but it can't hide? :laughing:

abayer:
ping @jglick - like I said on the old PR, what needs to be done for the APIs in question to be considered public?

jglick:
At a minimum, `WorkflowLibRepository.workspace()` could be `public` and have Javadoc explaining what you will find there, pointing you to the listener, etc.

abayer:
Okiedokie.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryListener.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
jglick: Sorry to be a pest but can you use the MIT license please, as specified in the POM?

abayer: Yipes, Intellij got silly. Fixing.

jglick:
:bee:

PR #2, opened 2016-05-06
[JENKINS-34650] Test for workflowLibs sandbox behavior

jglick:
[JENKINS-34650](https://issues.jenkins-ci.org/browse/JENKINS-34650)

Can reproduce problem but have no idea how to fix. Maybe @kohsuke knows. I tried

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
index fdbb0da..cf761cc 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsGroovyShell.java
@@ -5,6 +5,7 @@ import com.cloudbees.groovy.cps.NonCPS;
 import com.cloudbees.groovy.cps.SandboxCpsTransformer;
 import com.cloudbees.groovy.cps.TransformerConfiguration;
 import groovy.lang.Binding;
+import groovy.lang.GroovyClassLoader;
 import groovy.lang.GroovyCodeSource;
 import groovy.lang.GroovyShell;
 import groovy.lang.Script;
@@ -98,6 +99,15 @@ class CpsGroovyShell extends GroovyShell {
         return s;
     }
 
+    @Override
+    public GroovyClassLoader getClassLoader() {
+        CompilerConfiguration cc = new CompilerConfiguration();
+        CpsTransformer t = new CpsTransformer(); // TODO conditionally use SandboxCpsTransformer
+        t.setConfiguration(new TransformerConfiguration().withClosureType(CpsClosure2.class));
+        cc.addCompilationCustomizers(t);
+        return new GroovyClassLoader(super.getClassLoader(), cc);
+    }
+
     /**
      * Used internally to reload the script back when coming back from the persisted state
      * (therefore we don't want to record this.)
```

but then the resulting loader does not work at all:

```
[p #1] org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
[p #1] WorkflowScript: 1: unable to resolve class pkg.Privileged 
[p #1]  @ line 1, column 1.
[p #1]    new pkg.Privileged().write('direct-false')
[p #1]    ^
```

In the meantime, the test at least covers the control case.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.plugins.workflow.cps.global;
 
+import hudson.model.Result;
 import java.io.File;
 import java.util.Arrays;
 
@@ -16,7 +17,10 @@ import org.jvnet.hudson.test.RestartableJenkinsRule;
 import javax.inject.Inject;
 import org.apache.commons.io.FileUtils;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
+import org.junit.ClassRule;
 import org.junit.runners.model.Statement;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 
 /**
@@ -26,6 +30,7 @@ import org.jvnet.hudson.test.JenkinsRule;
  */
 public class WorkflowLibRepositoryTest {
 
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
     @Rule public RestartableJenkinsRule story = new RestartableJenkinsRule();
 
     @Inject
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
```diff
@@ -143,4 +148,55 @@ public void userDefinedGlobalVariable() throws Exception {
             }
         });
     }
+
+    /** Global libraries should run outside the sandbox, regardless of whether the caller is sandboxed. */
```
oleg-nenashev: IMHO such behavior should be configurable.

oleg-nenashev:
I have no idea how to fix it as well :( 
Maybe we need a kind of the "Pipeline stacktrace" support with markers ("Acquired lock", "Entered CPS library section", etc.)

jglick:
@kohsuke suggested using `CodeSource`.

abayer:
How would that be done?

jglick:
Well, if I knew, I would have done it already! :-)

abayer:
...fair enough =)

jglick:
@abayer or anyone else care to give me a bee? The scaffolding of the test is useful to have in place even without a fix.

abayer:
:bee:!

PR #7, opened 2016-06-29
[JENKINS-34517] Global variables caused ClassNotFoundException’s when resuming builds

jglick:
[JENKINS-34517](https://issues.jenkins-ci.org/browse/JENKINS-34517)

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
@@ -124,6 +124,9 @@ public class WorkflowLibRepositoryTest {
                         "  echo 'title was '+config.title",
                         "}")
                         , "\n"));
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryTest.java
```diff
@@ -124,6 +124,9 @@ public void userDefinedGlobalVariable() throws Exception {
                         "  echo 'title was '+config.title",
                         "}")
                         , "\n"));
+                FileUtils.writeStringToFile(new File(vars, "acmeClass.groovy"), StringUtils.join(Arrays.asList(
+                        "@groovy.transform.Field int answer = 42")
+                        , "\n"));
```
jglick: At @kohsuke’s suggestion, verifying some other previously untested idioms which we were concerned might get accidentally broken.

kohsuke:
:bee:

PR #9, opened 2016-08-09
[JENKINS-26192] Support @Grab from global libs

jglick:
[JENKINS-26192](https://issues.jenkins-ci.org/browse/JENKINS-26192)

@reviewbybees

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,3 +198,24 @@ jenkinsPlugin {
     name = 'git'
 }
 ```
+
+### Using third-party libraries
+
+You may use third-party Java libraries (typically found in Maven Central) within your library code via `@Grab`.
+Refer to the [Grape documentation](http://docs.groovy-lang.org/latest/html/documentation/grape.html#_quick_start) for details.
+For example:
+
+```groovy
+@Grab('org.apache.commons:commons-math3:3.4.1')
+import org.apache.commons.math3.primes.Primes
+void parallelize(int count) {
+  if (!Primes.isPrime(count)) {
+    error "${count} was not prime"
+  }
+  // …
+}
+```
+
+Since the _Overall/RunScripts_ permission is required to push any changes to the global library repository,
+library code is allowed to call any Java methods without sandbox checks.
+Therefore it can safely encapsulate calls to third-party libraries (as well as to Jenkins model APIs, etc.).
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.5</version>
-        <relativePath />
+        <version>2.12</version>
+        <relativePath/>
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-cps-global-lib</artifactId>
@@ -82,6 +82,11 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.apache.ivy</groupId>
+            <artifactId>ivy</artifactId>
+            <version>2.4.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
             <version>1.15</version>
@@ -91,13 +96,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>1.15</version>
+            <version>2.6-SNAPSHOT</version> <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/16 -->
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
-            <version>1.15</version>
+            <version>2.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.cps.global;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import groovy.grape.Grape;
 import groovy.grape.GrapeEngine;
 import hudson.init.InitMilestone;
@@ -39,12 +40,14 @@ public class GrapeHack {
 
     private static final Logger LOGGER = Logger.getLogger(GrapeHack.class.getName());
 
+    @SuppressFBWarnings(value="DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED", justification="the least of our concerns")
     @Initializer(after=InitMilestone.PLUGINS_PREPARED, fatal=false)
     public static void hack() throws Exception {
         String grapeIvyName = "groovy.grape.GrapeIvy";
+        String ivyGrabRecordName = "groovy.grape.IvyGrabRecord"; // another top-level class
         URL groovyJar = Grape.class.getProtectionDomain().getCodeSource().getLocation();
         LOGGER.log(Level.FINE, "using {0}", groovyJar);
-        ClassLoader l = new URLClassLoader(new URL[] {groovyJar}, new MaskingClassLoader(GrapeHack.class.getClassLoader(), grapeIvyName));
+        ClassLoader l = new URLClassLoader(new URL[] {groovyJar}, new MaskingClassLoader(GrapeHack.class.getClassLoader(), grapeIvyName, ivyGrabRecordName));
         Class<?> c = Class.forName(grapeIvyName, false, l);
         Field instance = Grape.class.getDeclaredField("instance");
         instance.setAccessible(true);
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -198,3 +198,24 @@ jenkinsPlugin {
     name = 'git'
 }
 ```
+
```
jglick: Click _Display the rich diff_ to see this formatted…

pom.xml
```diff
@@ -82,6 +82,11 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.apache.ivy</groupId>
+            <artifactId>ivy</artifactId>
+            <version>2.4.0</version>
```
jglick: The actual fix. Groovy includes Grape but `groovy-all` depends on the sole engine, based on Ivy, which is not otherwise in the classpath. Seems to suffice for it to be somewhere in the `uberClassLoader`.

pom.xml
```diff
@@ -91,13 +96,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>1.15</version>
+            <version>2.6-SNAPSHOT</version> <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/16 -->
```
jglick: downstream of https://github.com/jenkinsci/workflow-job-plugin/pull/16

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+
+@Issue("JENKINS-26192")
+public class GrapeTest {
+
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule public RestartableJenkinsRule story = new RestartableJenkinsRule();
+    @Inject private WorkflowLibRepository repo;
+
+    @Test public void useBinary() throws Exception {
```
jglick: the main use case

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -0,0 +1,174 @@
...
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p", WorkflowJob.class);
+                WorkflowRun b = p.getBuildByNumber(1);
+                SemaphoreStep.success("wait/1", null);
+                story.j.assertBuildStatusSuccess(story.j.waitForCompletion(b));
+                story.j.assertLogContains("got []", b);
+            }
+        });
+    }
+
+    @Test public void var() throws Exception {
```
jglick: or this one

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -0,0 +1,174 @@
...
+                p.setDefinition(new CpsFlowDefinition("pkg.X.run()", true));
+                story.j.assertLogContains("net.nowhere.Nonexistent", story.j.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0)));
+            }
+        });
+    }
+
+    // TODO test alternate Maven repositories
+
+    @Ignore("TODO MissingMethodException: No signature of method: static com.google.common.base.CharMatcher.whitespace() is applicable for argument types: () values: []")
+    @Test public void overrideCoreLibraries() throws Exception {
```
jglick: Do not see any way this could work, unfortunately, so you are in general pinned to the versions of libraries shipped in Jenkins.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -0,0 +1,174 @@
...
+        });
+    }
+
+    @Ignore("TODO works in Jenkins 2 / Groovy 2; in 1, fails with (concealed): unable to resolve class org.jenkinsci.plugins.docker.workflow.Docker")
+    @Test public void useSource() throws Exception {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                FileUtils.write(new File(repo.workspace, "src/pkg/Dokker.groovy"),
+                    "package pkg\n" +
+                    "@Grapes([@Grab('org.jenkins-ci.plugins:docker-workflow:1.7'), @Grab('org.jenkins-ci.plugins:docker-commons:1.3.1')])\n" +
```
jglick: Transitive dependencies do not seem to be working, not sure why.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -0,0 +1,174 @@
...
+                    "import org.jenkinsci.plugins.docker.workflow.Docker\n" +
+                    "static def stuff(script, body) {\n" +
+                    "  new Docker(script).node {body()}\n" +
```
jglick: Not a realistic use case, but I struggled to find a JAR in Central containing runnable Groovy source files—especially one accepting closure arguments (used here to prove that the library gets CPS-transformed).

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -0,0 +1,174 @@
...
+                    "import org.apache.commons.math3.complex.ComplexField\n" +
+                    "def call() {ComplexField.instance.one}");
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition("echo(/${one()} + ${one()} = ${one().add(one())}/)", true));
+                story.j.assertLogContains("(1.0, 0.0) + (1.0, 0.0) = (2.0, 0.0)", story.j.buildAndAssertSuccess(p));
+            }
+        });
+    }
+
+    // TODO test transitive dependencies; need to find something in Central which has a dependency not in this plugin’s test classpath and which could be used easily from a script
```
jglick: Suggestions welcome. I spent considerable time looking for a usable example, without success.

kohsuke:
:bee: I think this is really cool!!

jglick:
Should add a test verifying that an attempt to `@Grab` from a userland sandboxed script fails.

jglick:
Works in functional tests but not in actual usage; get a `NoClassDefFoundError: org/apache/ivy/core/settings/IvySettings` from `GrapeIvy.groovy:81`. Investigating.

jglick:
Afraid I see no way forward without core changes.

jglick:
Seems blocked by https://github.com/jenkinsci/jenkins/pull/2374 or the equivalent. Playing with options.

jglick:
I think I have it working now. Need to create an acceptance test so there is automated test coverage of it working in a realistic class loader environment.

jglick:
Note that this change makes `@Grab` work throughout the system, including in `/script` etc. I could do it more conservatively so that `Grape.instance` is only set while some CPS VM thread is executing, or perhaps even only while global libraries are being parsed during the start of a build, which would limit the damage if there is some unforeseen consequence.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
```diff
@@ -0,0 +1,70 @@
...
+            LOGGER.log(Level.FINE, "{0} was already set, not touching it", engine);
+        } catch (RuntimeException | LinkageError x) {
+            LOGGER.log(Level.FINE, "by default we could not load Grape", x);
+        }
+        String grapeIvyName = "groovy.grape.GrapeIvy";
+        String ivyGrabRecordName = "groovy.grape.IvyGrabRecord"; // another top-level class
+        URL groovyJar = Grape.class.getProtectionDomain().getCodeSource().getLocation();
+        LOGGER.log(Level.FINE, "using {0}", groovyJar);
+        ClassLoader l = new URLClassLoader(new URL[] {groovyJar}, new MaskingClassLoader(GrapeHack.class.getClassLoader(), grapeIvyName, ivyGrabRecordName));
+        Class<?> c = Class.forName(grapeIvyName, false, l);
```
jglick: Groovy includes Grape but `groovy-all` depends on the sole engine, based on Ivy, which is not otherwise in the classpath.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
```diff
@@ -0,0 +1,70 @@
...
+        Field instance = Grape.class.getDeclaredField("instance");
+        instance.setAccessible(true);
+        instance.set(null, c.newInstance());
```
jglick: Note that this makes `@Grab` work throughout the system, including in `/script` etc. I could do it more conservatively so that `Grape.instance` is only set while some CPS VM thread is executing, or perhaps even only while global libraries are being parsed during the start of a build, which would limit the damage if there is some unforeseen consequence.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
```diff
@@ -0,0 +1,70 @@
...
+    public static void hack() throws Exception {
+        try {
+            GrapeEngine engine = Grape.getInstance();
+            LOGGER.log(Level.FINE, "{0} was already set, not touching it", engine);
+        } catch (RuntimeException | LinkageError x) {
+            LOGGER.log(Level.FINE, "by default we could not load Grape", x);
+        }
+        String grapeIvyName = "groovy.grape.GrapeIvy";
+        String ivyGrabRecordName = "groovy.grape.IvyGrabRecord"; // another top-level class
+        URL groovyJar = Grape.class.getProtectionDomain().getCodeSource().getLocation();
```
kohsuke: I've generally used `Which.jarFile(Class)` to locate a jar file, and that contains various logics specific to various servlet containers. Have you considered that? This way of doing things is new to me.

In this case a failure is not fatal, so I suppose it's OK either way.

kohsuke:
:+1:

src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
kohsuke: I think this would be a feature that people will like. It should be documented somewhere.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/GrapeHack.java
jglick: That gives you a `File` via various hacks, but I just want a `URL`, under the assumption that whatever was good enough for the original `ClassLoader` should be good enough for `URLClassLoader` too. Yes that does assume that

* the original `ClassLoader` correctly associated `ProtectionDomain`s
* any custom URL protocol is properly registered via `URLStreamHandlerFactory`

Yes exotic containers might break either or both of those assumptions. Not sure what to do about that beyond testing (the downstream ATH test by default uses the built-in Jetty). `Which` does not really help because it is solving a different problem.

rsandell:
:bee:

PR #10, opened 2016-08-10
[JENKINS-31155] Shared libraries

jglick:
[JENKINS-31155](https://issues.jenkins-ci.org/browse/JENKINS-31155)

- [x] global configuration option
- [x] per-folder configuration option
- [X] `@Library` annotation
- [x] basic SCM checkout & load
- [x] test https://github.com/jenkinsci/git-plugin/pull/433
- [x] global variable support (incl. test of [JENKINS-32400](https://issues.jenkins-ci.org/browse/JENKINS-32400) scenario)
- [x] _Global Variables Reference_ support
- [x] Replay support
- [x] form validation for library definitions
- [x] resource loading
- [x] restart tests
- [x] `checkpoint` tests
- [x] `@Grab` integration test (presupposes #9)
- [x] basic interactive tests
- [x] dynamic GitHub `LibraryResolver`: https://github.com/jenkinsci/github-organization-folder-plugin/pull/22
- [x] interactive tests using an organization folder: https://github.com/cloudbeers/PR-demo/pull/15
- [x] ~~`library` step exploration~~ (seems like a dead end)
- [X] inline help (needs https://github.com/jenkinsci/scm-api-plugin/pull/13)
- [x] [documentation](https://github.com/jglick/workflow-cps-global-lib-plugin/blob/shared-libs-JENKINS-31155/README.md)
- [X] https://github.com/jenkinsci/subversion-plugin/pull/168
- [X] update Pipeline-as-code demo to use it: https://github.com/jenkinsci/github-branch-source-plugin/pull/72
- [x] refactor API to use `LibraryRetriever`
- [x] make sure every library requested is used
- [x] fix deletion of last entry in `GlobalLibraries`

@reviewbybees

PR #13, opened 2016-08-26
Added possibility to share groovy script over git repositories other than workflowLibs from Jenkins

attrobit:
Our goal was to load our shared groovy scripts from our own git repository. I extended the GroovyShellDecorator for that. The decorator parses the pipeline script, search for a specific pattern, download the git libraries and add them to the class path. The following example show a recognized pattern:

```
//#git http://<my git url> <branch>
```

I have to add some documentation and extend the GitLoader to support git credentials, but i want to know if you would accept my changes.


.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 work
+/nbproject/
```
jglick: Heh. Delete, you do not need it.

jglick:
Interesting work, but I guess you missed #10.

attrobit:
I saw #10 but it does not support our necessary feature of adding groovy files to the classloader. The library step are not part of it. Or how can i add groovy files from our own git repository to the groovy shell classloader?

jglick:
> it does not support our necessary feature of adding groovy files to the classloader

Sure it does, that is the whole point.

attrobit:
Looks great, thanks.

PR #15, opened 2016-09-08
[JENKINS-38048] Deleting obsolete query parameter

jglick:
Incidental to the real issue in [JENKINS-38048](https://issues.jenkins-ci.org/browse/JENKINS-38048) but should be fixed nonetheless.

@reviewbybees


rsandell:
:bee:

PR #16, opened 2016-10-04
[JENKINS-38550] Hide currently inapplicable LibraryRetrieverDescriptor’s

jglick:
[JENKINS-38550](https://issues.jenkins-ci.org/browse/JENKINS-38550)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java
@@ -37,6 +37,8 @@ import java.util.Collection;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/SCMRetriever.java
```diff
@@ -72,7 +74,7 @@ public SCM getScm() {
             return "Legacy SCM";
         }
 
-        public static List<SCMDescriptor<?>> getSCMDescriptors() {
+        public List<SCMDescriptor<?>> getSCMDescriptors() {
```
oleg-nenashev: 🐛 Binary compatibility change. Please keep [and deprecate] the original method and implement a new one if required

jglick: Will not do; this was for internal use only. I will mark it as restricted however.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java
```diff
@@ -116,6 +121,13 @@ public boolean isAllowVersionOverride() {
 
     @Extension public static class DescriptorImpl extends Descriptor<LibraryConfiguration> {
 
+        // TODO JENKINS-20020 ought to be unnecessary
+        public Collection<LibraryRetrieverDescriptor> getRetrieverDescriptors() {
```
daniel-beck: `@Restricted`?

svanoort:
Implementation LGTM but waiting for the annotations to block external consumption of the APIs here (as others noted) once done I can happily bee it.

svanoort:
also 🐝

daniel-beck:
Here's a bee for you: :bee:

oleg-nenashev:
Well, I'm approving a change since the maintainer feels strongly about the compat change. Not a bee

PR #17, opened 2016-10-04
[JENKINS-38712] Avoid deep indentation

jglick:
See [JENKINS-38712](https://issues.jenkins-ci.org/browse/JENKINS-38712) for why I did not do this to begin with: there are drawbacks to the new control.

jenkins-38712

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMRetriever.java
@@ -77,6 +77,8 @@ public class SCMRetriever extends LibraryRetriever {
         public List<SCMDescriptor<?>> getSCMDescriptors() {
             List<SCMDescriptor<?>> descriptors = new ArrayList<>();
             for (SCMDescriptor<?> d : SCM.all()) {
+                // TODO SCM._for cannot be used here since it requires an actual Job, where we want to check for applicability to Job.class
+                // (the best we could do is to check whether SCM.checkout(Run, …) is overridden)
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
🐝

recampbell:
There's not much of a visual separation between "Legacy SCM" and its apparent sub-options. They appear to be alternatives.

Perhaps if we added some help text (ie, the (?) icon) to this option and its sub option, it would help visual clarify the relationship of these radio boxes and also guide the user?

recampbell:
OK, I can't figure out how to edit a review. Was trying to leave a 🐛 to say that we should discuss this.

jglick:
Revised appearance with only *Legacy SCM*:

revised

Including also *Modern SCM*:

revised-with-modern

As I noted before, the original controls work better when you have sufficient horizontal space to display it, but we have no way that I know of to reduce the indentation.

jglick:
BTW it is also possible to use `f:block` to show a blurb unconditionally (in addition to or instead of the help button), as is done beneath section titles for *Pipeline Libraries* and *Global Pipeline Libraries*.

jglick:
I checked the appearance in 2.19.1 but it is no different, pending https://github.com/jenkinsci/cloudbees-folder-plugin/pull/67 and some equivalent for global configuration.

recampbell:
OK, I think this is a signficant improvement over what we have and directly addresses the raised issue. There's still some confusion since Legacy SCM looks like an alternative to Subversion, but this is a big step forward IMO and we should ship like this.

🐝

oleg-nenashev:
🐝

PR #18, opened 2016-10-10
[JENKINS-38517] Improvements to how libraries are checked out

jglick:
Noticed as part of [JENKINS-38517](https://issues.jenkins-ci.org/browse/JENKINS-38517), though the main thrust of that issue is really something else—that changing **Local module directory** should be discouraged.

@reviewbybees


abayer:
:bee:

PR #30, opened 2017-01-30
[JENKINS-39719] Integration test

jglick:
Demonstrates effectiveness of https://github.com/cloudbees/groovy-cps/pull/46. Downstream of https://github.com/jenkinsci/workflow-cps-plugin/issues/100.

@reviewbybees


jglick:
> triggering a config reload via JenkinsRule

Would not do anything to a `static` field.

PR #32, opened 2017-02-25
Added a Jenkinsfile, and fixed some buglets on Windows

jglick:
- [X] add something so that organization folder scanning knows about it
- [x] make sure tests pass on a new baseline
- [x] make sure tests pass on Windows

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,8 @@
         <jenkins.version>1.642.3</jenkins.version>
         <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
         <workflow-cps-plugin.version>2.28</workflow-cps-plugin.version>
+        <git-plugin.version>3.1.0</git-plugin.version>
+        <scm-api-plugin.version>2.1.1</scm-api-plugin.version>
     </properties>
     <dependencies>
         <dependency>
@@ -91,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
-            <version>1.3</version>
+            <version>${scm-api-plugin.version}</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -63,7 +63,10 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <jenkins-test-harness.version>2.19</jenkins-test-harness.version>
         <workflow-cps-plugin.version>2.28</workflow-cps-plugin.version>
+        <git-plugin.version>3.1.0</git-plugin.version>
```
svanoort: I really hate the use of a property here -- if you just move the normal and test deps adjacent it's trivial to bump both.  Use search+replace if you sometimes forget one.  This level of DRY is silly.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
```diff
@@ -69,7 +69,10 @@ public Object getValue(@Nonnull CpsScript script) throws Exception {
     public @CheckForNull String getHelpHtml() throws IOException {
         if (!help.exists())     return null;
 
-        return Jenkins.getActiveInstance().getMarkupFormatter().translate(FileUtils.readFileToString(help, Charsets.UTF_8));
+        return Jenkins.getActiveInstance().getMarkupFormatter().translate(
```
svanoort: Noice.

svanoort:
Needs to mention that you made a number of other small fixes in the title.

src/test/java/org/jenkinsci/plugins/workflow/libs/LibraryAdderTest.java
```diff
@@ -290,7 +292,7 @@ public static void register(Object o) {
             clearInvocationCaches.invoke(metaClass);
         }
         for (WeakReference<ClassLoader> loaderRef : LOADERS) {
-            MemoryAssert.assertGC(loaderRef);
+            MemoryAssert.assertGC(loaderRef, false);
```
svanoort: Aha!

jglick: Picked up from `workflow-cps` and verifies that there are no soft, much less hard, leaks.

PR #34, opened 2017-04-25
JENKINS-41497 make library changesets in a job configurable

hogarthj:
[JENKINS-41497](https://issues.jenkins-ci.org/browse/JENKINS-41497)

Although it can be useful to see what changes in a shared library could have an impact on a build there are many circumstances that shared library commits should be ignored.

This provides a configuration option to skip adding the changeset of a shared library to a build.

It defaults to including the changeset to match the current behaviour.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
@@ -42,12 +42,13 @@ public abstract class LibraryRetriever extends AbstractDescribableImpl<LibraryRe
      * Obtains library sources.
      * @param name the {@link LibraryConfiguration#getName}
      * @param version the version of the library, such as from {@link LibraryConfiguration#getDefaultVersion} or an override
+     * @param changesets whether to include changesets in the library in jobs using it from {@link LibraryConfiguration#getIncludeInChangesets}
      * @param target a directory in which to check out sources; should create {@code src/**}{@code /*.groovy} and/or {@code vars/*.groovy}, and optionally also {@code resources/}
      * @param run a build which will use the library
      * @param listener a way to report progress
      * @throws Exception if there is any problem (use {@link AbortException} for user errors)
      */
-    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
+    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull boolean changesets, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
 
     /**
      * Offer to validate a proposed {@code version} for {@link #retrieve}.
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
@@ -51,6 +51,17 @@ public abstract class LibraryRetriever extends AbstractDescribableImpl<LibraryRe
     public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull boolean changesets, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
 
     /**
+     * Obtains library sources.
+     * @param name the {@link LibraryConfiguration#getName}
+     * @param version the version of the library, such as from {@link LibraryConfiguration#getDefaultVersion} or an override
+     * @param target a directory in which to check out sources; should create {@code src/**}{@code /*.groovy} and/or {@code vars/*.groovy}, and optionally also {@code resources/}
+     * @param run a build which will use the library
+     * @param listener a way to report progress
+     * @throws Exception if there is any problem (use {@link AbortException} for user errors)
+     */
+    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
+
+    /**
      * Offer to validate a proposed {@code version} for {@link #retrieve}.
      * @param name the proposed library name
      * @param version a proposed version
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
@@ -42,13 +42,13 @@ public abstract class LibraryRetriever extends AbstractDescribableImpl<LibraryRe
      * Obtains library sources.
      * @param name the {@link LibraryConfiguration#getName}
      * @param version the version of the library, such as from {@link LibraryConfiguration#getDefaultVersion} or an override
-     * @param changesets whether to include changesets in the library in jobs using it from {@link LibraryConfiguration#getIncludeInChangesets}
+     * @param changelog whether to include changesets in the library in jobs using it from {@link LibraryConfiguration#getIncludeInChangesets}
      * @param target a directory in which to check out sources; should create {@code src/**}{@code /*.groovy} and/or {@code vars/*.groovy}, and optionally also {@code resources/}
      * @param run a build which will use the library
      * @param listener a way to report progress
      * @throws Exception if there is any problem (use {@link AbortException} for user errors)
      */
-    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull boolean changesets, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
+    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull boolean changelog, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
 
     /**
      * Obtains library sources.
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
@@ -85,6 +85,7 @@ public class LibraryStep extends AbstractStepImpl {
     private static final Logger LOGGER = Logger.getLogger(LibraryStep.class.getName());
 
     private final String identifier;
+    private Boolean changelog = true;
     private LibraryRetriever retriever;
 
     @DataBoundConstructor public LibraryStep(String identifier) {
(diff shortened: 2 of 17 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRetriever.java
```diff
@@ -42,12 +42,13 @@
...
      * @param name the {@link LibraryConfiguration#getName}
      * @param version the version of the library, such as from {@link LibraryConfiguration#getDefaultVersion} or an override
+     * @param changesets whether to include changesets in the library in jobs using it from {@link LibraryConfiguration#getIncludeInChangesets}
      * @param target a directory in which to check out sources; should create {@code src/**}{@code /*.groovy} and/or {@code vars/*.groovy}, and optionally also {@code resources/}
      * @param run a build which will use the library
      * @param listener a way to report progress
      * @throws Exception if there is any problem (use {@link AbortException} for user errors)
      */
-    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
+    public abstract void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull boolean changesets, @Nonnull FilePath target, @Nonnull Run<?,?> run, @Nonnull TaskListener listener) throws Exception;
```
jglick: Incompatible.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
```diff
@@ -154,6 +154,7 @@ public AutoCompletionCandidates doAutoCompleteIdentifier(@AncestorInPath ItemGro
             String[] parsed = LibraryAdder.parse(step.identifier);
             String name = parsed[0], version = parsed[1];
             boolean trusted = false;
+            boolean changesets = true;
```
jglick: This should really be an argument to the step, not an aspect of the library.

src/test/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetrieverTest.java
```diff
@@ -66,4 +69,59 @@
         }
     }
 
+    @Test public void includeChanges() throws Exception {
```
jglick: `@Issue`

hogarthj:
Thanks for the review, I'm working on the bits you raised now.
As for the UX I think it important that at the config level when the shared library is described it's important to be able to set this as otherwise implicit loaded libraries would not be able to have the changesets hidden. 
I do agree that ultimately the developer writing their Jenkinsfile ought to be able to make the choice.
What I'm working on right now is extending @Library and the library step to allow for this, and then the config checkbox just becomes the default behaviour for the shared library.
The admin sets the default (important for implicit libraries) but would allow the person writing the Jenkisnfile to make the ultimate choice.

hogarthj:
@jglick The configuration tick box for including changesets or not is now to set the default behaviour for that shared library, but anyone writing a script can override that setting by adding @(changesets|nochangesets|true|false|yes|no) to the end of their library identifier.

hogarthj:
@jglick do you have any other comments that I need to work on to incorporate or are you happy with this default -> override configuration option?

We'd really like to get this rolled out to our workplace, since as we add more shared library capabilities we're getting frequent questions about "confusing and random recent changes" on the jobs of various teams.

hogarthj:
@jglick your thoughts? We'd really like to get this in place to avoid the changelogs from our central share libraries from appearing in recent changes for our developers, without the trouble of effectively using our own fork of this plugin

rogerwelin:
@jglick ping. our org is also very interested in this change

ghenzler:
@jglick Our Organisation also is struggling with this - using this checkbox seems to be the most-straight-forward solution (and where I would have expected it). Big 👍 for this.

hogarthj:
@jglick it's fast approaching a month since the PR was updated in response to your review requests ...

can you please provide an update on your thoughts, or if it matches the requests in your review to then merge it?

jglick:
> otherwise implicit loaded libraries would not be able to have the changesets hidden

Not true; you can always override an implicit load with an explicit `@Library`.

hogarthj:
> Not true; you can always override an implicit load with an explicit @Library

But then it wouldn't be implicitly loaded as it'd be an explicit declaration which would go against the concept of the global shared library from a central CICD team maintaining the toolsets that development teams make use of without having to add @Library and know of the library name involved, what branch it is etc etc 

When you said:

> @Library is a bit trickier but probably the annotation could have a new member

My understanding was to modify the way the string is split (ie change the annotation) to include this setting, as the lightest touch possible that "just worked" with both the `@Library` annotation and the step... since they both use the same string parser.

I'll take a step back again and look at how `@Library` could be extended with another param instead.

Thanks for taking a look at this and I'll get it updated to match your thoughts.

jglick:
> then it wouldn't be implicitly loaded as it'd be an explicit declaration

Right; only necessary if you wish to override the changelog setting, just like you would do so today if you wished to override the version.

> My understanding was to modify the way the string is split

The `value` attribute. Possible, but just makes for a parsing mess. Previously I had suggested

```java
boolean changelog() default true;
```

which is more explicit. (Perhaps the version should have been a separate attribute, too, but too late for that now.)

hogarthj:
I *think* we are talking the same thing but cross purposes to an extent ...

Right now the "default" (well only, but you get the idea) option is to include changes to any shared libraries loaded in the recent changes of a job, but with no way to override it.

What I'd like to see is the ability to set the default to not include changes to any shared libraries loaded in the recent changes, and that's primarily what this was initially about and the first PR code was focused on with the global config setting.

I do agree that the name@version string parsing in retrospect is a bit messy ... 

Would you consider an iteration on this where one can write:
```
@Library([[name: 'foo', version: 'bar', changelog: false], ['something@devel]])
```
so that it is backwards compatible to avoid breaking code (ie not touch the existing string parser for `value` at all) but going forwards can provide more flexibility in a cleaner manner?

Whether you'd want to do a formal deprecation of the existing syntax is up to you of course, but I don't think it's too late to make a stepped change if it helps for future maintenance and removes a 'parsing hack' which assumes things about the '@'  

I'll start updating the PR in this direction at any rate until I hear otherwise from you :)

jglick:
Well simply

```groovy
@Library(value='name@master', changelog=false)
```

would suffice, and should be straightforward since you are merely adding a new annotation method.

jglick:
(not sure if Groovy supports `java.lang.annotation.Repeatable`; if not, need `@Libraries`)

hogarthj:
@jglick the way this works now pretty much matches the discussion we've had, there's still a few rough edges but I'd appreciate your thoughts ... 

I've not touched the `@Library(["foo", "bar", "woo"])` multiple library syntax but those will follow whatever the default ticked in the global configuration page is.

To override the global configuration settings screen using the library annotation `@Library(value="foo@version", changelog=true|false)` will set the behaviour the way the developer writing the Jenkinsfile wants it to be.

Using the `library identifier: "foo", retriever: <SCMTHING>, changelog: false` syntax allows to disable including the library changelog in a step, but it defaults to including to match the current behaviour.

There's test cases that cover all these scenarios.

I wanted to make the library step behaviour the same as `@Library` with changelog just acting as an override to the global configuration, but I've been struggling with the LIbraryStep configRoundtrip test and 
```
LibraryStep s = new LibraryStep("foo");
r.assertEqualDataBoundBeans(s, stepTester.configRoundTrip(s));
```
not passing with allowing the databoundsetter changelog to be null but still defaulting to including the global configuration setting as the changelog property then doesn't match going in to the test as null but coming out as true ...

Is this at least on the path you'd expect? I'd love to get to a "minimal viable product" state for a merge to mitigate the issues several people have discussed here and on jira.

marcospedreiro:
My team is responsible for managing our central CI/CD pipelines, and we are able to add a lot of value for all of our other developers with the implicitly loaded global libraries. However the issue that this PR set out to fix has limited our ability to add to the global library, since we do not want to unnecessarily trigger jobs whenever we update the global library. 

I understand that there is value in the option to:

```groovy
@Library(value='name@master', changelog=false)
```

However ensuring that all teams and developers at a large organization add that line to their scripts is unmanageable for a single, small team to do. Allowing us to check a configuration option in the global library settings to be able to implicitly load the global library _and_ not trigger builds when we update the master branch would be a huge win. Furthermore, having one option does not necessarily preclude having the other.

hogarthj:
@marcospedreiro the issue this PR set out to fix is different to your issue.

This is concerned around a build including changes to a shared library included being listed in the "recent changes" properties of a build.

The original intention was just to do this at a global level, but with the guidance of @jglick this PR also allows the behaviour to be defined in the Jenkinsfile, overriding the global value.

The issue with setting a poll and that also picking up shared library commits is a separate one.

In the issue I tested a workaround of:
`For now I suggest using the workaround on that with the "skip if path affected" or "skip based on message" set to a wildcard of everything.`

That did, in my testing, avoid a poll being set on the SCM for the job also causing a trigger from commits to the shared libraries.

maxfields2000:
"skip if path affect" or "skip if message" do not work if there is not workspace. The git method will be forced to run the job to create the workspace.  

Regardless. Polling should not fire because a shared library changes, that should not require custom features, it should be a default behavior with the option to allow it to be included)

mslusarczyk:
@hogarthj your PR seems to fix issue that @marcospedreiro is having by accident. That's because of this [line](https://github.com/jenkinsci/workflow-scm-step-plugin/blob/17a9946bfe3f89df875d2d419d13aacb927467df/src/main/java/org/jenkinsci/plugins/workflow/steps/scm/SCMStep.java#L120). Poll is already [set to false](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/2c67d0554567d36856bebdb46961cfa869ca483a/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java#L90) on this `GenericSCMStep` so changing `changelog` prevents builds from being triggered by lib change.

rogerwelin:
@jglick what's the status of this PR? Would really like to implement the functionality

ghenzler:
@jglick We are waiting for this too (especially for the global option to disable change sets and triggers, we do not need this configurable in the `@Library` syntax)

sams-gleb:
We'd like to have this global option to disable changesets as well (currently we using our patch for global-lib plugin to exclude them). From my point of view, we shouldn't have a library changeset in each build for each project at all, because we usually come across a lot of projects which will use the same library and thus we seeing the same changeset everywhere  - that's a bit confusing

hogarthj:
@abayer The place I needed this for I've actually left and moved onto a new contract ... but I'd be happy to either see you finish it off or to pick it up with final polish given some direction.

I thought about defaulting to not including the changeset as that's what I'd *prefer* but I figured from a minimal changes perspective that was likely to pass muster and need the least debate to get through to keep the behaviour that is current the same, unless someone actively set the global config to disable it or added the changeset=false in their library stanzas.

abayer:
Oh riiiight, forgot about the global config option. With that, yeah, I think retaining the previous behavior as the default is acceptable for now, at least.

Lemme poke at a few things and make sure everything else is good to go, but I think I'll be merging this today.

abayer:
Yup, this is good. Merging now, checking to see if there's anything else in flight that should land before a release.

pauxus:
While this works like a charm, the actual `Library` annotation class should also be changed to include the new attribute for it to be valid. In our case, we are running all our libaries through automated tests (compiling and running agains mocked script instances). Currently this fails because of compilation errors (no such member).

PR #35, opened 2017-05-03
Add programmatic configuration for plugin

technolo-g:
This change adds a groovy script that can be used in the
Jenkins init system to configure the plugin without
using the UI.

@jglick not sure if this is the right place for this (or if you folks have interest). Let me know your thoughts :)


jglick:
This would be more like a blog post I suppose. In general you can use `init.groovy` to define system-level configuration, or create a `$JENKINS_HOME` “base image” with some `*.xml` configuration predefined, or in the future `system-config-dsl` would be a clearer approach.

PR #36, opened 2017-05-06
[JENKINS-41157] Allow modifications to trusted libraries

timja2:
[JENKINS-41157](https://issues.jenkins-ci.org/browse/JENKINS-41157)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -31,7 +31,6 @@ import hudson.FilePath;
 import hudson.model.Queue;
 import hudson.model.Run;
 import hudson.model.TaskListener;
-import hudson.model.User;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
@@ -270,8 +269,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
 
     private static boolean currentUserHasRunScriptsPermission() {
         return Jenkins.getActiveInstance()
-                .getAuthorizationStrategy()
-                .getACL(User.current())
+                .getACL()
                 .hasPermission(Jenkins.RUN_SCRIPTS);
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -31,6 +31,7 @@ import hudson.FilePath;
 import hudson.model.Queue;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import hudson.model.User;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


timja-kainos:
I've added the tests, but I'm not sure how to make jenkins authenticate in a test currently so it looks like either security is disabled or running with ADMINISTER atm.
Any pointers appreciated.

timja-kainos:
I've added tests that look right to me but they are failing, will debug when I get more time

timja:
I've added tests that look right to me but they are failing, will debug when I get more time

timja:
Ready for review

timja:
@reviewbybees ?

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -266,6 +268,13 @@
 
     }
 
+    private static boolean currentUserHasRunScriptsPermission() {
+        return Jenkins.getActiveInstance()
+                .getAuthorizationStrategy()
+                .getACL(User.current())
```
jtnord: this to me looks incorrect; it returns the ACL for the `User` object - which the current user will likely always have full control of given that it is the current users `User`.

Should check `Jenkins.getActiveInstance().getACL().hasPermission(Jenkins.RUN_SCRIPTS)`

timja-kainos:
on cloudbees sorry wasn't sure what the procedure was for being the PR to attention.

timja:
on cloudbees sorry wasn't sure what the procedure was for being the PR to attention.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
timja: It did work, but I've made the change to the other method

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -266,6 +267,12 @@
 
     }
 
+    private static boolean currentUserHasRunScriptsPermission() {
+        return Jenkins.getActiveInstance()
+                .getACL()
+                .hasPermission(Jenkins.RUN_SCRIPTS);
```
amuniz: This checks permissions on current user, which IIRC is `SYSTEM` when a build is automatically triggered (i.e. react to a webhook call). So I think this would be transitively by-passing the trusted flag for any user filing a PR.

src/test/java/org/jenkinsci/plugins/workflow/libs/LibraryAdderTest.java
```diff
@@ -245,23 +252,57 @@
...
+                        "", "*", "", true)))));
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("@Library('trusted@master') import trusted; trusted()", true));
+        final WorkflowRun b1 = r.buildAndAssertSuccess(p);
+        r.assertLogContains(originalMessage, b1);
+
+        ACL.impersonate(User.get("bob").impersonate(), new Runnable() {
+            @Override public void run() {
+                ReplayAction ra = b1.getAction(ReplayAction.class);
+                assertEquals(Collections.emptyMap(), ra.getOriginalLoadedScripts());
```
jglick: This does not prove anything interesting, only that the untrusted user cannot use the web form to mount an attack.

jglick:
FYI this code is now in https://github.com/jenkinsci/pipeline-groovy-lib-plugin.

PR #37, opened 2017-05-11
Implement a workspace retriever.

iocanel:
I wanted to validate pull requests send against my pipeline library. So I needed to be able to dynamically load the library in the pull request branch.

With the current retrievers, that is not possible _(but I may miss something so please, correct me if I am wrong)_. So my best bet was to `checkout scm` and try to load the library from the workspace.  To make it work I hacked a simple WorkspaceRetriever and thought it might make sense to share...

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceRetriever.java
@@ -1,0 +1,74 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.workflow.libs;
+
+import org.jenkinsci.Symbol;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.io.IOException;
+
+import javax.annotation.Nonnull;
+
+import hudson.Extension;
+import hudson.FilePath;
+import hudson.model.Computer;
+import hudson.model.Node;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import jenkins.model.Jenkins;
+
+public class WorkspaceRetriever extends LibraryRetriever {
+
+    private final String sourcePath;
+
+    @DataBoundConstructor
+    public WorkspaceRetriever(String sourcePath) {
+        this.sourcePath = sourcePath;
+    }
+
+    @Override
+    public void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull FilePath target, @Nonnull Run<?, ?> run, @Nonnull TaskListener listener) throws Exception {
+        Node node = Jenkins.getActiveInstance();
+
+        Computer computer = node.toComputer();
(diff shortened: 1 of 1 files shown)


jglick:
Not sure what it is you are trying to accomplish here (there is no test which might demonstrate its usage, or help file, or JIRA issue), but probably you just wanted to use the `library` step which allows you to specify a library location.

iocanel:
I think that what best demonstrates my use case is this:

https://github.com/syndesisio/syndesis-pipeline-library/blob/master/Jenkinsfile#L3

For pull request validation shake, I wanted to be able to checkout the library from the pull request and load it as is. As I don't have an easy way of referring to the pull request branch, my only shot was to try and checkout the source and load it from my workspace.

This pull request is adding the retriever that allows me to do so.

jglick:
More concisely:

```groovy
library identifier: "syndesis-pipeline-library@$BRANCH_NAME", retriever: workspaceRetriever(WORKSPACE)
```

src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceRetriever.java
```diff
@@ -0,0 +1,74 @@
...
+    private final String sourcePath;
+
+    @DataBoundConstructor
+    public WorkspaceRetriever(String sourcePath) {
+        this.sourcePath = sourcePath;
+    }
+
+    @Override
+    public void retrieve(@Nonnull String name, @Nonnull String version, @Nonnull FilePath target, @Nonnull Run<?, ?> run, @Nonnull TaskListener listener) throws Exception {
+        Node node = Jenkins.getActiveInstance();
```
jglick: This will not work when using an agent. Would require a new API in `LibraryRetriever` allowing the `LibraryStep` to pass the `StepContext.get(FilePath.class)` if one exists (in general this may return null).

jglick:
Perhaps instead try just

```groovy
library identifier: '...', retriever: legacySCM(scm)
```

I think that would address your use case, assuming you were not patching the library after checkout. Needs testing of course. CC @rtyler

jglick:
(note: no `@` in the `identifier`)

Actually I think that would not work as is, but probably a retriever based on the concept could. TBD

jglick:
@jtnord / @batmat could my trick above with `legacySCM(scm)` simplify the testing system you described [here](https://jenkinsworld20162017.sched.com/event/AVaz/jumping-on-the-continuous-delivery-bandwagon-from-100-freestyle-jobs-to-pipelines-tactics-pitfalls-woes)?

jtnord:
@jglick I'm not sure if/where it would help at all.  Where you thinking of something specific (e.g. the `script.replace("checkout scm", "checkout git://.....git")`

/cc @batmat

schnatterer:
I have been waiting for a feature like this for a long time!
However, my use case is different, that's why I like to mention it here FYI:

I've experienced that for complex builds the `Jenkinsfile` gets kind of bulky and not very maintainable.
Now that the build job is code, it would be wonderful to have the same means as for other code. That is, I would like to divide it into smaller (more maintainable units) and unit test them.

See also [my question at SO](https://stackoverflow.com/q/46213913/1845976).

Looking forward to using the `workspaceRetriever`!

diegorusso:
Folks, any ETA on this?

tworec:
Gentlemen,
will it be pushed forward. I think it is vital option for shared pipeline libs.

iocanel:
As this covers my needs as is, and I am really tight on schedule, I don't intend to do additional work anytime soon. If anyone wants to push work on this PR please do.

tworec:
well, how do you use this code then?
Tests do not pass and findbugs complains. How do you build and deploy this on your instance?

samifruit514:
This feature would help a lot for local development. I could just mount my project lib folder in docker-compose and change my files directly

mfuchs:
Would this have security implications for PRs?

For example, it is possible to configure a multibranch job so that the Jenkinsfile for a PR is only used in the PR job for github users with Admin/Write rights.
I suppose the same should also be true for a local pipeline library.

jglick:
I believe the trick demonstrated in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43 supersedes the need for this.

schnatterer:
@jglick the trick works in general. However, it's not possible to load libs from subdirectories.
Comparing the questions vs answer upvotes of the [Stackoverflow question](https://stackoverflow.com/q/46213913/1845976) cited in #43, I presume it's not only me who looks forward to this feature 😉

jglick:
> it's not possible to load libs from subdirectories

#40 probably.

schnatterer:
TBH I can't judge if #37 or #40 would be the better solution from a technical point of view.

Neither 
* PR #37 (which has been dangling here for almost 2 years) nor
* PR #40 (the author closed the issue behind it in the meantime  - [JENKINS-46721](https://issues.jenkins-ci.org/browse/JENKINS-46721))

look very promising.

From a users perspective I'd like to load shared libs from a subdirectory within a repo.
Any chance one of this gets merged eventually?

heyleke:
See #99

src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceRetriever.java
Ramakrishna-Mandepudi: May be https://plugins.jenkins.io/workflow-cps-global-lib-http/ is worth trying.

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b workspace-retriever
git pull https://github.com/iocanel/workflow-cps-global-lib-plugin workspace-retriever
```

iocanel:
Stale, closing

schnatterer:
@iocanel have you found a solution or workaround?

iocanel:
> @iocanel have you found a solution or workaround? 

No, I used a fork with this fix, till my job shifted to completely different areas.

PR #40, opened 2017-09-07
JENKINS-46721: Expect library sources in custom subdirectory

petr-motejlek:
[JENKINS-46721](https://issues.jenkins-ci.org/browse/JENKINS-46721)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -105,6 +105,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
             boolean kindTrusted = kind.isTrusted();
             for (LibraryConfiguration cfg : kind.forJob(build.getParent(), libraryVersions)) {
                 String name = cfg.getName();
+                String sourceDir = cfg.getSourceDir();
                 if (!cfg.isImplicit() && !libraryVersions.containsKey(name)) {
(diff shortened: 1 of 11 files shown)


petr-motejlek:
Oh. There's a failed build. But after examining it, it couldn't have failed because of changes in this PR. It failed because it could not download a build of the credentials plugin.

basil:
Nice! My company has the same problem, and I'd really like for this functionality to be available.

basil:
Any update on this pull request? It would be nice to have this feature available.

jglick:
Also best to address [JENKINS-44481](https://issues.jenkins-ci.org/browse/JENKINS-44481) in the same change.

adamvoss:
@jglick I'm really just looking for the ability to load a library from a directory within the current workspace ([similar StackOverflow question](https://stackoverflow.com/q/46213913/1072626)).  Since my workspace comes from a SCM repository, this PR would address that need in a roundabout way so I have been looking into this.

However, with that being my use case (which SO shows to be shared by others) is this really the right solution?  It seems like it would be better to have a way to specify I actually want to load from the file system rather than saying to do a new SCM (Git) clone and then look at a sub-directory.

> Also best to address [JENKINS-44481](https://issues.jenkins-ci.org/browse/JENKINS-44481) in the same change.

I commented on that issue.  I'm not entirely sure if it applies or is desirable here.

dwnusbaum:
Closing in favor of https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/99.

PR #41, opened 2017-09-18
[JENKINS-33358] Test against 2.73.x since that includes Groovy 2.4.11

jglick:
Like https://github.com/jenkinsci/workflow-cps-plugin/pull/175.

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.32.3', '2.47'])
+buildPlugin(jenkinsVersions: [null, '2.32.3', '2.60.3', '2.73.1'])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.32.3', '2.60.3', '2.73.1'])
+buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'])
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.31</version>
+        <version>2.33</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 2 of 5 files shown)


Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.32.3', '2.47'])
+buildPlugin(jenkinsVersions: [null, '2.32.3', '2.60.3', '2.73.1'])
```
jglick: Probably you want to bump up the `jenkins.version` to 2.7.3, in which case we could drop the tests against 2.32.x I think.

Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.32.3', '2.47'])
+buildPlugin(jenkinsVersions: [null, '2.60.3', '2.73.1'])
```
jglick: 2.7.x–2.47 uses one version of Groovy, 2.48–2.60.x another, 2.61+ yet another. At least we are dropping 1.x!

pom.xml
```diff
@@ -62,8 +62,8 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
-        <workflow-cps-plugin.version>2.30</workflow-cps-plugin.version>
+        <jenkins.version>2.7.3</jenkins.version>
+        <workflow-cps-plugin.version>2.39</workflow-cps-plugin.version>
```
jglick: last security advisory

pom.xml
```diff
@@ -108,7 +108,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.27</version>
+            <version>1.33</version>
```
jglick: ditto, plus some follow-up fixes likely to affect testing of Groovy-heavy code like this

src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java
```diff
@@ -93,7 +93,6 @@
 
     // TODO test transitive dependencies; need to find something in Central which has a dependency not in this plugin’s test classpath and which could be used easily from a script
 
-    @Ignore("TODO ClassNodeResolver.tryAsLoaderClassOrScript swallows the CompilationFailedException and throws a generic GroovyBugError; fixed in Jenkins 2 / Groovy 2")
```
jglick: Can finally have automated tests cover Jenkins 2 behaviors.

pom.xml
abayer: We could theoretically go straight to 1.34, but 1.33 is good enough, yeah.

jglick:
Webhooks broken pending https://github.com/kohsuke/github-api/pull/373, will try to retrigger.

jglick:
@reviewbybees done

PR #43, opened 2018-03-12
Demonstrating that retriever: legacySCM(scm) can be used to self-test libraries

jglick:
Compare to #37. [This trick](https://stackoverflow.com/a/49112612/12916) apparently works, so…we ought to have a test verifying that it keeps working.

@reviewbybees @schnatterer

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.2</version>
+        <version>3.6</version>
         <relativePath />
     </parent>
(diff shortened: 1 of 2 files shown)


schnatterer:
Good point, @jglick!

I'd like to link #40 here, as I think it is closely related.
The use case in the [Stackoverflow question](https://stackoverflow.com/q/46213913/1845976) mentioned in this PR acutally is more about the subdirectory issue.

Still, `legacySCM(scm)` form self-testing of libraries is important feature for the development shared libraries.
See [here](https://github.com/cloudogu/ces-build-lib/blob/b1f01a1ec4ec1dc581870d4b8d71b3e3d8ba87af/Jenkinsfile) for a shared lib example.
Thank you, @jglick!

PR #44, opened 2018-04-06
Test adding tool from pipeline library

olivergondza:
I have got this test written desperately trying to reproduce an unrelated problem.


olivergondza:
Merging presuming noone would mind adding more tests.

PR #45, opened 2018-06-29
Allow shared libraries to be loaded from repo workspace

wesmcouch:
The purpose of this PR is to solve [JENKINS-46721](https://issues.jenkins-ci.org/browse/JENKINS-46721) - Expect library sources in custom subdirectory

This PR is a combination of ideas based on the work done in https://github.com/karolgil/SharedLibrary , https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/40 and https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/37

The WorkspaceLibraryAdder checks to see if any paths specified in ```@Library()``` exist in the workspace. If they exist those paths are added to the classpath of the non-trusted shell.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceLibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceLibraryAdder.java
@@ -11,8 +11,19 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 
+/**
+ * Adds groovy files in the workspace
+ * of the repository loaded to the classpath of the build
+ */
 public class WorkspaceLibraryAdder extends ClasspathAdder {
 
+    /**
+     * @param execution a running build (possibly newly started, possibly resumed)
+     * @param libraries aggregated entries from all encountered {@link Library#value} (will be empty if {@link Library} is never used at all); an implementation should remove entries it “claims”
+     * @param changelogs not used when adding directories
+     * @return Returns a list of directories that were added to the classpath
+     * @throws Exception
+     */
     @Nonnull
     @Override
     public List<Addition> add(@Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception {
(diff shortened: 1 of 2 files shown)


wesmcouch:
@varju

wesmcouch:
@karolgil @petr-motejlek @iocanel

varju:
Would it be useful to provide an example here of how this would be used, along with a proposed addition to https://jenkins.io/doc/book/pipeline/shared-libraries/?

karolgil:
Love it, looks like a nice addition. I agree that we need a working example of how to use it in Pipelines code.

wesmcouch:
Thanks for the suggestion, I have created an [example of the shared library here.](https://github.com/wesmcouch/workspace-level-shared-lib-doc) I wasn't able to find if the Jenkins documentation is in a public github repository, so if anyone has a link to that I can open a PR there too. If not the linked repo here can be added to the shared library documentation, and the example can be added to any example repository.

doctorseus:
This might be helpful but I don't think this addresses JENKINS-46721 in the right way as now you need to do changes in all your pipeline scripts in order to make a non-standard shared library structure work which is defined and configured in a complete different location.

IMHO https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/40#pullrequestreview-96534557 in addition to this still makes sense.

src/main/java/org/jenkinsci/plugins/workflow/libs/WorkspaceLibraryAdder.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Nonnull
+    @Override
+    public List<Addition> add(@Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception {
+        List<Addition> additions = new ArrayList<>();
+
+        Iterator<String> iterator = libraries.iterator();
+        while (iterator.hasNext()) {
+            String path = getWorkspaceLibraryPath(execution, iterator.next());
+            File libraryPath = new File(path);
```
jglick: This is not going to work for builds on agents.

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b feature/workspace-library
git pull https://github.com/wesmcouch/workflow-cps-global-lib-plugin feature/workspace-library
```

PR #48, opened 2018-07-16
[JENKINS-52313] Add encoding argument to ResourceStep and support Base64 encoding

dwnusbaum:
See [JENKINS-52313](https://issues.jenkins-ci.org/browse/JENKINS-52313).

Adds support for a special encoding named "Base64" that can be used for binary files. Useful in tandem with jenkinsci/workflow-basic-steps-plugin#66 to copy binary resource files from shared libraries to the workspace.

~I'd prefer to use Java 8's Base64 API but the plugin is currently building against Java 7 so I am using commons-compress via Jenkins core.~ Baseline has been updated to Java 8 and Jenkins 2.60.3 to use the new Base64 API.

@reviewbybees @jennbriden

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -35,6 +35,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -47,7 +48,6 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
-import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 import org.jenkinsci.plugins.workflow.cps.GlobalVariable;
@@ -215,7 +215,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
 
     private static String readResource(FilePath file, @CheckForNull String encoding) throws IOException, InterruptedException {
         if ("Base64".equals(encoding)) {
-            return Base64.encodeBase64String(IOUtils.toByteArray(file.read()));
+            return Base64.getEncoder().encodeToString(IOUtils.toByteArray(file.read()));
         } else {
             return IOUtils.toString(file.read(), encoding);
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -45,7 +45,10 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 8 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -202,14 +205,22 @@
...
             }
         }
         return resources;
     }
 
+    private static String readResource(FilePath file, @CheckForNull String encoding) throws IOException, InterruptedException {
+        if ("Base64".equals(encoding)) {
+            return Base64.encodeBase64String(IOUtils.toByteArray(file.read()));
+        } else {
+            return IOUtils.toString(file.read(), encoding);
```
svanoort: Worth mentioning that the downstream method is null-safe, which is how we handle null encodings. 

@jglick WDYT wrt the handling of on-master vs. on-build-agent encoding here?

svanoort: Worth including a javadoc explaining handling of null/empty (i.e.: null uses default encoding for this system, and empty string is coerced to null).

src/main/java/org/jenkinsci/plugins/workflow/libs/ResourceStep.java
```diff
@@ -50,6 +53,14 @@ public String getResource() {
         return resource;
     }
 
+    public String getEncoding() {
```
svanoort: `@CheckForNull` due to legacy encodings.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
dwnusbaum: Ok, I also need to mention the Base64 encoding in the docs.

src/main/java/org/jenkinsci/plugins/workflow/libs/ResourceStep.java
```diff
@@ -50,6 +53,14 @@ public String getResource() {
...
+        return encoding;
+    }
+
+    @DataBoundSetter public void setEncoding(String encoding) {
```
svanoort: Would be helpful to have javadoc explaining handling of null & empty strings here.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
svanoort: This maintains current behavior where we use the default encoding, so I think it is OK from a compatibility perspective.

src/main/resources/org/jenkinsci/plugins/workflow/libs/ResourceStep/config.jelly
```diff
@@ -28,4 +28,7 @@ THE SOFTWARE.
     <f:entry field="resource" title="Resource path">
         <f:textbox/>
     </f:entry>
+    <f:entry field="encoding" title="${%Character encoding}">
```
dwnusbaum: I completely forgot to add this before, but thankfully caught it when updating the documentation.

dwnusbaum:
`GrapeTest#var` is failing with `java.lang.RuntimeException: Error grabbing Grapes -- [download failed: org.apache.commons#commons-math3;3.4.1!commons-math3.jar]`. I can't reproduce it locally, so maybe it's just a transient error? I will try to retrigger the build to see if anything changes.

src/main/resources/org/jenkinsci/plugins/workflow/libs/ResourceStep/config.jelly
svanoort: Probably because it's something that *ought* to have been there before.

PR #49, opened 2018-07-25
Allowing a library name to be passed to the libraryResource step

jmcshane:
This would allow Jenkinsfile authors to specify the library name that is included to avoid duplication across resources.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/ResourceStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/ResourceStep.java
@@ -34,6 +34,7 @@ import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
 import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousStepExecution;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * Step to load a resource from a library.
@@ -41,6 +42,7 @@ import org.kohsuke.stapler.DataBoundConstructor;
 public class ResourceStep extends AbstractStepImpl {
 
     private final String resource;
+    private String libraryName;
 
     @DataBoundConstructor public ResourceStep(String resource) {
         this.resource = resource;
@@ -48,6 +50,15 @@ public class ResourceStep extends AbstractStepImpl {
 
     public String getResource() {
         return resource;
+    }
+    
+    public String getLibraryName() {
+        return libraryName;
+    }
+    
+    @DataBoundSetter
+    public void setLibraryName(String libraryName) {
+        this.libraryName = libraryName;
     }
 
     @Extension public static class DescriptorImpl extends AbstractStepDescriptorImpl {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/ResourceStep.java
```diff
@@ -74,10 +85,13 @@ public DescriptorImpl() {
 
         @Override protected String run() throws Exception {
             String resource = step.resource;
+            String libraryName = step.libraryName;
             Map<String,String> contents = LibraryAdder.findResources((CpsFlowExecution) getContext().get(FlowExecution.class), resource);
-            if (contents.isEmpty()) {
+            if (libraryName != null && contents.containsKey(libraryName)) {
+                return contents.get(libraryName);
+            } else if (contents.isEmpty() || (libraryName != null && !contents.containsKey(libraryName))) {
```
dwnusbaum: I think it would be better to have different error message for the `libraryName != null && !contents.containsKey(libraryName))` condition saying the resource was found, but not in the specified library.

src/main/java/org/jenkinsci/plugins/workflow/libs/ResourceStep.java
```diff
@@ -49,6 +51,15 @@
...
         return resource;
     }
+    
+    public String getLibraryName() {
+        return libraryName;
+    }
+    
+    @DataBoundSetter
+    public void setLibraryName(String libraryName) {
+        this.libraryName = libraryName;
```
dwnusbaum: I would use `= Util.fixEmptyAndTrim(libraryName)` to remove surrounding whitespace from `libraryName`.

dwnusbaum:
I just merged #48 which introduced a merge conflict, sorry! Also, it would be great to add a `help-libraryName.html` file explaining how to use the new parameter (similarly to what was done for the encoding parameter in #48).

jmcshane:
I should be able to get around to these comments today, I’ll make sure to rebase and resolve that conflict.

jmcshane:
@dwnusbaum Could you check now? Thanks!

src/main/resources/org/jenkinsci/plugins/workflow/libs/Messages.properties
```diff
@@ -23,3 +23,4 @@
 LibraryDecorator.could_not_find_any_definition_of_librari=Could not find any definition of libraries {0}
 ResourceStep.library_resource_ambiguous_among_librari=Library resource {0} ambiguous among libraries {1}
 ResourceStep.no_such_library_resource_could_be_found_=No such library resource {0} could be found.
+ResourceStep.no_matching_resource_found_in_library=Library resource {0} not found in library {1}, but was present in other library
```
dwnusbaum: nit: Maybe reword to: "Library resource {0} not found in library {1}, although it was present in libraries {2}" and pass `contents.keySet()` as the third argument.

src/main/resources/org/jenkinsci/plugins/workflow/libs/ResourceStep/help-libraryName.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    The shared library name to use to locate the resource.
+    If left blank, all loaded shared libraries will be loaded.
```
dwnusbaum: nit: Maybe reword to "If left blank, then this step will search for the resource in all configured shared libraries."

jglick:
Tread carefully here. I am not convinced this is safe, which is why I did not offer such an option to begin with.

jglick:
Also from a usage perspective this is dubious. `libraryResource` is normally called from library code. But a library does not, in general, know its own name—this is picked arbitrarily by an administrator. So you could not use this feature _portably_. It could only be used by a library written for exclusive use in a particular environment. But then why would there be duplication?

It would be helpful to have an explanation of what the original problem was which you are attempting to solve via this feature.

caporegim:
Any progress with this pull request?

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b SELECT-BRANCH-NAME
git pull https://github.com/jmcshane/workflow-cps-global-lib-plugin master
```

PR #50, opened 2018-08-20
[JENKINS-38992] Allow caching library versions

julienduchesne:
We have a jenkins instance that need to pull a lot of libraries. When many jobs are running simultaneously it can sometimes take up to 30 seconds to load a single library.

Probably fixes https://issues.jenkins-ci.org/browse/JENKINS-38992 and https://issues.jenkins-ci.org/browse/JENKINS-48210 in some measure


julienduchesne:
Alright, and I still have to fix failing tests. FYI, that's how it looks in the UI 
image

julienduchesne:
Added tests, the build failure seems unrelated:
```java.lang.RuntimeException: Error grabbing Grapes -- [download failed: org.apache.commons#commons-math3;3.4.1!commons-math3.jar]```

**Comment by julienduchesne** [comment]
Any update on this?

**Comment by julienduchesne** [comment]
Hey @dwnusbaum. I've fixed every issue except putting the config in SCMSourceRetriever. I put the config in LibraryConfiguration on purpose, that way it is independant from the SCM. The same way that other configs such as the default version and default version override are. I also don't think that it's a bug if I change the repository but the version stays the same. Let me know what you think.

File: src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java [hunk 1]
```diff
@@ -119,12 +117,20 @@ public boolean getIncludeInChangesets() {
...
+        this.cachingConfiguration = cachingConfiguration;
+    }
+
     @Nonnull boolean defaultedChangelogs(@CheckForNull Boolean changelog) throws AbortException {
-      if (changelog == null) {
-        return includeInChangesets;
-      } else {
-        return changelog;
-      }
+        if (changelog == null) {
```
jglick: Please avoid touching lines of text with only whitespace changes, so as to keep diffs minimal. You can configure your text editor to refrain from doing so.

src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration/help-excludedVersionsStr.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    Space separated list of versions excluded from caching. Ex: "master release10 release9"
```
jglick: But there is no possibility here to specify a regular expression.

FWIW my proposal in JENKINS-38992 was to make this automatic—zero configuration. The SCM should know whether or not a given version was deterministic. The current `SCMSource.retrieve(String, TaskListener)` does not suffice since at least in the case of `AbstractGitSCMSource` the resulting `SCMRevision.isDeterministic` will always be true—the method resolves the actual current head of a branch. Thus we would need to extend the `SCMHead` API in `scm-api` to let the SCM source indicate whether the _head_ is deterministic (like a hash, or arguably a tag) vs. nondeterministic (like a branch). This would be easy to implement in `git` at least.

julienduchesne: I agree that it should support regex and maybe add an include field as well (only include a single version) but is it really user friendly to cache some versions but not others. In our use case, it would be annoying to cache tags by default because we use semantic versioning and publish minor versions as well as patch versions (ex: release v4.2.1 also releases v4.2 and v4).

All I'm saying is I'm not a fan of hidden behavior and the current implementation is very simple. Is it really necessary to complicate it?

jennbriden:
I have had several customers report hitting performance issues due to shared libraries not being cached. What LTS do we think this will go into?

julienduchesne:
It works very well though. We are using it on a jenkins instance running about 20 very short jobs (15 sec to 5 minutes) concurrently at all times during the day. We used to only be able to do a full cycle of our workflow in over an hour before this. It now takes under 20 minutes.

criley-eagle:
> "Based on input here, this might not be the best implementation."
@jennbriden Is this pr still viable, or is there a better implementation in the works, as we are definitely hitting up against the same issue this pr was attempting to resolve?

julienduchesne:
I'm keeping this updated in case you guys ever want to merge it :). It's been working fine without a single glitch since I started using it and our job load has about doubled since then.

julienduchesne:
Would you guys be willing to take another look at this? It might not (arguably) be the absolute best solution but it works flawlessly and it is the most simple. How can we get this merged?

ppattard:
+1 
Would love to have this story delivered. We are heavily relying on pipeline libraries and fetching them at each build is costly + introduces a point of failure on the repository we fetch from.

ajrussellsap:
I think this is a really important improvement that would improve build times for a lot of people.  Even if the implementation might not be 100% perfect, the benefits outweigh any downsides.  Please can we get this merged?

kmurrayAtCloudbees:
Is there a plan to merge this anytime soon?

rpannek:
We would need some solution for caching libraries too because we're moving our from our internal infra to github because we're open sourcing it.

julienduchesne:
This PR is battle-tested by now. We have run over 200k jobs with it (No exaggeration, I archived most of build history not too long ago). Honestly, without this, Jenkins would just be unusable for us. What changes would be needed for this to be merged?

Inetov:
@bitwiseman @dwnusbaum guys, can you help with the review?

MarkEWaite:
@oleg-nenashev the [CI job is failing on ci.jenkins.io](https://ci.jenkins.io/job/Plugins/job/workflow-cps-global-lib-plugin/job/PR-50/) because the JGit version provided in git plugin 3.12.1 cannot merge the changes from this branch with the current master branch.  

The version of JGit included in the latest release of the git plugin (4.1.1) and git client plugin (3.1.1) **can merge** the two branches.  Command line git **can merge** the two branches as well.  The merge failure is specific to JGit 4.7.0.

The original submitters could resolve the issue by merging the master branch to their branch.  We could resolve the issue by updating the plugins on ci.jenkins.io.  I'll check with @olblak on the policy for plugin updates on ci.jenkins.io.

patrickhuy:
@bitwiseman @dwnusbaum @oleg-nenashev it seems like this would give great improvements to a lot of Jenkins users. Is there any chance on moving this forward?

oleg-nenashev:
I pinged the maintainers multiple times over past months. Unfortunately I cannot  do much more about it, but I will try pinging again

julienduchesne:
:wave: I haven't received any "pings" until now. What needs to be done to have this merged?

msrb:
@oleg-nenashev LGTM? :)

oleg-nenashev:
> 👋 I haven't received any "pings" until now. What needs to be done to have this merged?

Looks like somebody needs to take ownership of the plugin or to get the current maintainers to review/merge the pull request. Unfortunately I cannot do much more to push it forward at the moment.

atmcarmo:
Thanks for this PR! Hope this gets merged. It's really an issue for us the time it gets to clone and load a library when many pipelines are starting to execute at the same time.

robodenitro:
any updates here?

MarkEWaite:
@robodenitro I've been running this pull request since early November 2020 without detecting any issue.  You could do the same by installing the file from https://github.com/MarkEWaite/docker-lfs/commit/408684e046549fb3b8539ec5ded85006fdde11ab .  Once you do that, your specific need should be resolved and you can report your test results

It is more persuasive to report successful use of a pull request than to ask for a status update.

nnadams:
Following Mark's example, I have been using this pull request since June 2020 and seen no issues. We run thousands of jobs weekly that all make use of this caching. Our CI has been faster and much more robust. I hope this can be merged. Either way thank you @julienduchesne!

src/test/java/org/jenkinsci/plugins/workflow/libs/ResourceStepTest.java
```diff
@@ -53,18 +57,87 @@
     @Rule public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
 
     @Test public void smokes() throws Exception {
-        sampleRepo.init();
-        sampleRepo.write("src/pkg/Stuff.groovy", "package pkg; class Stuff {static def contents(script) {script.libraryResource 'pkg/file'}}");
-        sampleRepo.write("resources/pkg/file", "fixed contents");
-        sampleRepo.git("add", "src", "resources");
-        sampleRepo.git("commit", "--message=init");
+        initFixedContentLibrary();
+
```
MarkEWaite, line 61: Remove trailing white space
```suggestion

```

src/test/java/org/jenkinsci/plugins/workflow/libs/ResourceStepTest.java
```diff
@@ -53,18 +57,87 @@
...
+        WorkflowRun secondBuild = r.buildAndAssertSuccess(p);
+        r.assertLogContains("got fixed contents", secondBuild);
+        r.assertLogContains("Library stuff@master is cached. Copying from home.", secondBuild);
+        r.assertLogNotContains("git", secondBuild); // git is not called
+    }
+
+    @Test public void cachingExcludedLibrary() throws Exception {
+        clearCache("stuff");
+        initFixedContentLibrary();
+
```
MarkEWaite, line 93: Remove trailing white space
```suggestion

```

src/test/java/org/jenkinsci/plugins/workflow/libs/ResourceStepTest.java
```diff
@@ -53,18 +57,87 @@
...
             new LibraryConfiguration("stuff", new SCMSourceRetriever(new GitSCMSource(null, sampleRepo.toString(), "", "*", "", true)))));
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
         p.setDefinition(new CpsFlowDefinition("@Library('stuff@master') import pkg.Stuff; echo(/got ${Stuff.contents(this)}/)", true));
         r.assertLogContains("got fixed contents", r.buildAndAssertSuccess(p));
     }
 
+    @Test public void caching() throws Exception {
+        clearCache("stuff");
+        initFixedContentLibrary();
+
```
MarkEWaite, line 72: Remove trailing white space
```suggestion

```

src/test/java/org/jenkinsci/plugins/workflow/libs/ResourceStepTest.java
```diff
@@ -53,18 +57,87 @@
...
+        WorkflowRun secondBuild = r.buildAndAssertSuccess(p);
+        r.assertLogContains("got fixed contents", secondBuild);
+        r.assertLogContains("Library stuff@other is excluded from caching.", secondBuild);
+        r.assertLogContains("git", secondBuild); // git is called
+    }
+
+    @Test public void cachingRefresh() throws Exception {
+        clearCache("stuff");
+        initFixedContentLibrary();
+
```
MarkEWaite, line 115: Remove trailing white space
```suggestion

```

MarkEWaite (approved):
Remove trailing white space in diffs

ababushk:
We use version of workflow-cps-global-lib with changes from this PR on production instance with huge amount of jobs and builds per day. No issues so far!
Moreover, this allowed us to run multiple concurrent builds (>1000 simultaneously) without waiting for a lock on git repo cache directory for our global library on Jenkins server, which has speeded up them drastically. Many thanks to @julienduchesne !

atmcarmo:
This solution seems to be rock solid, really improving performance even with a very big amount of builds happening at the same time.

What needs to be done to merge to be included in a new release?

Thank you very much for your work 👏

michelgasser:
Why is the PR not being merged? Why is it taking so long?

timja:
@dwnusbaum @car-roll @bitwiseman Would it be possible for someone to take a look at this?

oleg-nenashev:
As a CloudBees employee I spent so much time pinging the current
maintainers. Doing the last attempt

car-roll (approved):
So, I'm just going over the comments here and I think this summarizes what went on here:

It does look like most issues were addressed, but there was a preference by both @dwnusbaum and @jglick to implement things slightly different, which is why this PR stalled. But because there was no significant effort to address the alternatives proposed by them, and since this PR was the ONLY solution available, it became widely adopted. Thus the wide adoption of this solution is now causing pressure for this PR to be merged. 

Given that this PR has been actively used in the wild now for quite some time, I'm going ahead to merge it.

michelgasser:
Cool, thanks a lot for the merge! :-)

MaximeAnsquer:
How can I disable caching please?

MaximeAnsquer:
Thank you @MarkEWaite!

aaronr93:
Hi @MarkEWaite, just following up on my first question above. Is there a way to programmatically clear the cache from a pipeline?

MarkEWaite:
> Hi @MarkEWaite, just following up on my first question above. Is there a way to programmatically clear the cache from a pipeline?

No method that I know to clear cache from a Pipeline.

PR #51, opened 2018-09-08
[JENKINS-52313] fix to a file leak introduced in libraryResource

tschaible:
This wraps access to the InputStream of a file with a try-with-resources block.  Prior to recent changes, access to the file contents was done with readToString which does the same.

This change was causing a pretty severe file leak, and this update will wrap the resource in a try-with-resources block like the original implementation to prevent the leak.


tschaible:
I opened this with the wrong ticket number, i will fix that and re-open.

PR #52, opened 2018-09-08
[JENKINS-53485] fix to a file leak introduced in libraryResource

tschaible:
See [JENKINS-53485](https://issues.jenkins-ci.org/browse/JENKINS-53485). 

This wraps access to the InputStream of a file with a try-with-resources block. Prior to recent changes, access to the file contents was done with readToString which does the same.

This change was causing a pretty severe file leak, and this update will wrap the resource in a try-with-resources block like the original implementation to prevent the leak.


tschaible:
Thanks for the quick turnaround and all your contributions.  Jenkins Pipelines are awesome.

PR #53, opened 2018-09-08
[JENKINS-40109] Pre-evaluate compilation errors to make them serializable

rudolfwalter:
One way to solve [JENKINS-40109](https://issues.jenkins-ci.org/browse/JENKINS-40109).

I have one small uncertainty: do you think anyone could be relying on the details inside the MultipleCompilationErrorsException? Should we save it as a transient field inside CompilationErrorException (and override getCause() to return it)?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
@@ -12,7 +12,14 @@ import java.io.NotSerializableException;
  */
 public class CompilationErrorException extends RuntimeException {
     public CompilationErrorException(MultipleCompilationErrorsException original) {
-        super(original.toString());
+        super(original.getMessage());
         setStackTrace(original.getStackTrace());
     }
+
+    @Override
+    public synchronized Throwable fillInStackTrace() {
+        return this;
+    }
+
+    public static final long serialVersionUID = 1;
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
@@ -1,0 +1,18 @@
+package org.jenkinsci.plugins.workflow.cps.global;
+
+import org.codehaus.groovy.control.MultipleCompilationErrorsException;
+
+import java.io.NotSerializableException;
+
+/**
+ * An exception that replaces Groovy's {@link MultipleCompilationErrorsException},
+ * because that is not serializable (which would, under certain circumstances,
+ * lead to the user being presented with a {@link NotSerializableException} instead
+ * of a compilation error -- see JENKINS-40109).
+ */
+public class CompilationErrorException extends RuntimeException {
+    public CompilationErrorException(MultipleCompilationErrorsException original) {
+        super(original.toString());
+        setStackTrace(original.getStackTrace());
+    }
+}
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
@@ -1,0 +1,42 @@
+package org.jenkinsci.plugins.workflow.cps.global;
+
+import com.cloudbees.hudson.plugins.folder.Folder;
+import hudson.model.Run;
+import hudson.plugins.git.GitSCM;
+import jenkins.plugins.git.GitSCMSource;
+import jenkins.plugins.git.GitSampleRepoRule;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.libs.GlobalLibraries;
+import org.jenkinsci.plugins.workflow.libs.LibraryConfiguration;
+import org.jenkinsci.plugins.workflow.libs.SCMRetriever;
+import org.jenkinsci.plugins.workflow.libs.SCMSourceRetriever;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.util.Collections;
+
+@Issue("JENKINS-40109")
+public class CompilationErrorExceptionTest extends Assert {
+    @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
+    @Rule public JenkinsRule r = new JenkinsRule();
+    @Rule public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
+
+    @Test public void catchSyntaxError() throws Exception {
+        sampleRepo.init();
+        sampleRepo.write("vars/mymagic.groovy", "def call() { echo([a:'alpha' b:'beta']) } //syntax error: missing comma");
+        sampleRepo.git("add", "vars");
+        sampleRepo.git("commit", "--message=init");
+        GlobalLibraries.get().setLibraries(Collections.singletonList(new LibraryConfiguration("badlib", new SCMSourceRetriever(new GitSCMSource(sampleRepo.toString())))));
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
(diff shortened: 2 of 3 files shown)


rudolfwalter:
Pinging @abayer (assignee of the ticket)

src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.io.NotSerializableException;
+
+/**
+ * An exception that replaces Groovy's {@link MultipleCompilationErrorsException},
+ * because that is not serializable (which would, under certain circumstances,
+ * lead to the user being presented with a {@link NotSerializableException} instead
+ * of a compilation error -- see JENKINS-40109).
+ */
+public class CompilationErrorException extends RuntimeException {
```
dwnusbaum: Probably good to go ahead and add an explicit `serialVersionUID`.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
```diff
@@ -0,0 +1,18 @@
...
+    public CompilationErrorException(MultipleCompilationErrorsException original) {
+        super(original.toString());
```
dwnusbaum: nit: I'd have to check what the exact output is to know for sure, but my first guess is that it would be better to use `original.getMessage()` instead of `original.toString()`.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
```diff
@@ -0,0 +1,18 @@
...
+        setStackTrace(original.getStackTrace());
```
dwnusbaum: The call to `super()` will call `fillInStackTrace` to create a stack trace for the new exception, and since you are setting the stack trace explicitly here I think you should override `fillInStackTrace` so that it just returns `this` and doesn't do any actual work.

dwnusbaum:
> do you think anyone could be relying on the details inside the MultipleCompilationErrorsException

I'm not sure. Maybe our exception could extend [CompilationFailedException](http://docs.groovy-lang.org/latest/html/api/org/codehaus/groovy/control/CompilationFailedException.html) or its superclass instead so we could attach more of the information from the original exception, but if none of that shows up in the message in the first place then I doubt anyone is relying on it.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
rudolfwalter: That is what I meant to call, except I mistyped last second. Nice catch.

rudolfwalter:
I'll write a test to showcase the problem.

Also, we cannot extend CompilationFailedException because it has non-serializable members.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Rule public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
+
+    @Test public void catchSyntaxError() throws Exception {
+        sampleRepo.init();
+        sampleRepo.write("vars/mymagic.groovy", "def call() { echo([a:'alpha' b:'beta']) } //syntax error: missing comma");
+        sampleRepo.git("add", "vars");
+        sampleRepo.git("commit", "--message=init");
+        GlobalLibraries.get().setLibraries(Collections.singletonList(new LibraryConfiguration("badlib", new SCMSourceRetriever(new GitSCMSource(sampleRepo.toString())))));
+        WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("@Library('badlib@master') _; try { mymagic(); echo 'why are we here?' } catch(err) { echo 'forcing CPS serialization...'; sleep 1; echo 'did we die?' }", true));
```
dwnusbaum: nit: Might be helpful to use `echo err.getMessage()` in the first `catch` block and then `r.assertLogContains("syntax error")` to make it obvious what error is expected to be thrown there.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import java.util.Collections;
+
+@Issue("JENKINS-40109")
+public class CompilationErrorExceptionTest extends Assert {
```
dwnusbaum: nit: Maybe better to just create a file `UserDefinedGlobalVariable.java` in case we want to add more tests for that class, but I think things are fine as-is.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
rudolfwalter: Done.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
rudolfwalter: I say, let's leave it as is. It can always be renamed later if needed.

dwnusbaum:
Tested this out locally today to verify that without the fix the new test fails with a `NotSerializableException`, which it did. The CI build appears to be failing because of an infra-related issue, and the tests pass for me locally.

src/test/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorExceptionTest.java
```diff
@@ -0,0 +1,43 @@
...
```
abayer: Why are you extending `Assert` here?

dwnusbaum: Probably just a copy-paste from elsewhere in the plugin (https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/2c83e4735d4fa65e60e04db095d2918ba44cfd5b/src/test/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariableListTest.java#L24), but yeah it would be good to clean it up.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/CompilationErrorException.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.NotSerializableException;
+
+/**
+ * An exception that replaces Groovy's {@link MultipleCompilationErrorsException},
+ * because that is not serializable (which would, under certain circumstances,
+ * lead to the user being presented with a {@link NotSerializableException} instead
+ * of a compilation error -- see JENKINS-40109).
+ */
+public class CompilationErrorException extends RuntimeException {
```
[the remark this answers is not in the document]
dwnusbaum: Yeah that approach makes sense to me. I'm happy to do this bit if @rudolfwalter is fine with it.

dwnusbaum:
> How will this work for a compilation failure in src?

I did some quick testing, and it looks like there are 2 cases:
* When using `@Library` and `import`ing the classes with syntax errors, the compilation errors in `src` are thrown straight out of `WorkflowRun#run`, so we don't need to worry about the exception being serialized as part of the program in that case. (Although we could still convert the exception for consistency.)
* When using the `library` step, the compilation errors in `src` are thrown out of `LibraryStep$LoadedClasses#loadClass`, so we'll need to add a catch clause for `MultipleCompilationErrorsException` there as well.

I'll open a PR in workflow-cps to move the exception and a new one in this repo on top of this PR to handle the `src` cases as well.

rudolfwalter:
@dwnusbaum Should I make the changes @abayer requested, or will your PR replace this one completely and you'll do them there?

dwnusbaum:
@rudolfwalter My plan was to file a new PR in both places, but the new PR here would still have all of your commits (so you'll still have attribution on them), but if you'd rather do it yourself, that's totally fine too!

rudolfwalter:
I just wanted to clarify if there was anything else I needed to do. Go right ahead! :)

dwnusbaum:
@rudolfwalter Nope you're good, thanks for getting this started! :)

dwnusbaum:
Closing in favor of #54. Thanks @rudolfwalter!

PR #54, opened 2018-10-01
[JENKINS-40109] Make compilation errors serializable

dwnusbaum:
See [JENKINS-40109](https://issues.jenkins-ci.org/browse/JENKINS-40109). Subsumes #53 (CC @rudolfwalter).

This PR uses the exception introduced in jenkinsci/workflow-cps-plugin#250 to make compilation errors thrown during shared library loading serializable. The PR also updates some dependencies to satisfy the Maven enforcer and incrementalifies the plugin.

@reviewbybees


Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.60.3', '2.89'])
+buildPlugin(jenkinsVersions: [null, '2.138.1'])
```
dwnusbaum: Current baseline is 2.60.3, so there's no point to test against it twice, and I figured 2.89 was probably added to test against newer Jenkins versions so I figured it made sense to update to the latest LTS.

dwnusbaum:
Test failures are legitimate, probably caused by the dependency updates. I'm looking into them.

abayer:
@dwnusbaum It's the core version - needs to be `2.62` or higher thanks to `workflow-cps`.

dwnusbaum:
@abayer Yeah I saw that, and tested out bumping it, but still had some issues locally (probably because I didn't clean out build artifacts correctly before updating it). I am rerunning tests locally against 2.73.x.

dwnusbaum:
Looks like a test unrelated to this change is failing because the path is too long for Windows: 

> stderr: C:/Jenkins/workspace/cps-global-lib-plugin_PR-54-NAGP2QJUN4SJP4A3KASYPBGZI57HSOSNKF5LUE7P53LG3GKFD6KQ/target/tmp/jenkinsTests.tmp/jenkins6108956040851236483test/workspace/mp_branch-TTIP6J74P423JEJOPRQJUYPNRMKWAGJEYWO2IWTIU6S3CBQSL2UQ@script/.git/refs: Filename too long

I'll take a look and see what I can do to shorten it.

src/test/java/org/jenkinsci/plugins/workflow/libs/SCMRetrieverTest.java
```diff
@@ -49,6 +51,7 @@
 
     @Url("https://stackoverflow.com/a/49112612/12916")
     @Test public void selfTestLibraries() throws Exception {
+        Assume.assumeFalse("File names are too long for Windows", Functions.isWindows());
```
dwnusbaum: It doesn't feel right to try to nickel and dime the file names to save a few characters ("mp" -> "m", "branch" -> "b", "master" -> "m"), when the main issue (I think) is due to having 2 hashes included from branch-api in the path (One for ci.jenkins.io, and one for the test itself) in addition to the integer suffix on the JenkinsRule's JENKINS_HOME directory. AIUI jenkinsci/branch-api-plugin#129 might make this better.

jglick: https://github.com/jenkinsci/plugin-pom/pull/125 may help somewhat.

jglick: 3.24 now

dwnusbaum: Still failing, will probably just revert and leave the test ignored for now on Windows. It seems strange that there are nested `junit` folders in the workspace, but I'm not sure if that is related to any of the recent changes: `...\target\tmp\junit5127255811482076289\junit7031506459549069085`

Jenkinsfile
svanoort: 🤷‍♂️

PR #55, opened 2018-10-04
[JENKINS-48516] Uniquify distinct groups of radio buttons

dwnusbaum:
See [JENKINS-48516](https://issues.jenkins-ci.org/browse/JENKINS-48516).

Right now, if you have multiple shared libraries defined, when you load the configuration page, only the last library's SCM settings will be visible. 

The reason for this seems to be that the names of radio button groups are not distinct for each library. [repeatable.js](https://github.com/jenkinsci/jenkins/blob/22aa2e6e766074d11249893e3f35e0b99e20d3d0/core/src/main/resources/lib/form/repeatable/repeatable.js#L189) tries to add a prefix to uniquify radio buttons in distinct `repeatable-chunk`s, but here we have ambiguity even within a single `repeatable-chunk`, and I'm not sure if the `repeatable.js` code works correctly when loading existing data vs adding new radio buttons to a live page dynamically using `repeatable:add` buttons (Based on [the generated HTML](https://gist.github.com/dwnusbaum/746e0a7a4d2cab9e6f28733178c51f47) I'd think that there should be no issue, so it seems like the JS-based uniquifying might happen too late to fix dupes during the initial page load).

I'd like to upstream this fix into core (EDIT: jenkinsci/jenkins#3672) for all `f:descriptorRadioList`s, since I don't think it would ever hurt, but I am opening this PR independently since I assume we won't want to bump our core dependency to the bleeding edge for a fix like this.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
@@ -40,5 +40,5 @@ THE SOFTWARE.
     <f:entry field="includeInChangesets" title="${%Include @Library changes in job recent changes}">
         <f:checkbox default="true"/>
     </f:entry>
-    <f:descriptorRadioList varName="retriever" instance="${instance.retriever}" descriptors="${descriptor.retrieverDescriptors}" title="${%Retrieval method}"/>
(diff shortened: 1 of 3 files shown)


dwnusbaum:
Currently trying to create an HtmlUnit test for this bug, but am having issues getting it to fail before the change. If I can't get a regression test working by tomorrow I will probably give up on the test.

dwnusbaum:
Given that the fix was backported to the 2.138 LTS line in 2.138.3, I don't think it's worth adding this workaround here. Users on 2.138.3 or higher will get the bugfix automatically.

PR #60, opened 2019-01-25
Switch from entry to optional block

jsoref:
See https://issues.jenkins-ci.org/browse/JENKINS-55787

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
@@ -31,14 +31,8 @@ THE SOFTWARE.
     <f:entry field="defaultVersion" title="${%Default version}">
         <f:textbox/>
     </f:entry>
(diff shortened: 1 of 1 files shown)


jsoref:
Please hold off on this, I need to work on the migration bits :-(.

dwnusbaum:
No problem, good to see that we have some tests to catch regressions in this area! If you do get back to it, before/after screenshots in the context of shared library configuration would be great!

jsoref:
I think what's going to happen is...
Given current code:
```
<f:entry title="...">
  <f:checkbox.../>
</f:entry>
```
with current proposed replacement:
```
<f:optionalBlock title="..." ...>
```
new proposal:
```
<f:entry>
  <f:checkbox  title="...".../>
</f:entry>
```

The question then is: given that the PR branch is "optional-block" and the new change doesn't involve `optionalBlock`, should I close all my PRs and make new ones? :-/

(I'm still running tests on all of the plugins I'm changing, and ones those basic tests finish running, I get to reinstall all of the plugins into my two test jenkins instances -- normal and "sleek".)

PR #61, opened 2019-01-28
Clarify that you can control if Pipeline shared library changes should…

rkivisto:
… trigger new builds for consuming jobs

It's not currently obvious in the Jenkins UI that the `Include @Library changes in job recent changes` option not only will control if the Pipelines will list the library changes in their change log, but it also controls whether a new build would be triggered whenever the Pipeline shared library is updated for any job that includes the library.

I'm proposing updating the description of the checkbox as well as the help text to make this more clear.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
@@ -37,7 +37,7 @@ THE SOFTWARE.
     <f:entry field="allowVersionOverride" title="${%Allow default version to be overridden}">
         <f:checkbox default="true"/>
     </f:entry>
-    <f:entry field="includeInChangesets" title="${%Include @Library changes in job recent changes, and changes to this library should trigger jobs that use it}">
+    <f:entry field="includeInChangesets" title="${%Include @Library changes in job recent changes}">
         <f:checkbox default="true"/>
     </f:entry>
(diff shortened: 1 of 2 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration/config.jelly
```diff
@@ -37,7 +37,7 @@ THE SOFTWARE.
     <f:entry field="allowVersionOverride" title="${%Allow default version to be overridden}">
         <f:checkbox default="true"/>
     </f:entry>
-    <f:entry field="includeInChangesets" title="${%Include @Library changes in job recent changes}">
+    <f:entry field="includeInChangesets" title="${%Include @Library changes in job recent changes, and changes to this library should trigger jobs that use it}">
```
dwnusbaum: Would be good to check how this looks in the UI.

dwnusbaum: Ok, yeah I think this part of the change should be reverted, since it makes things align way too far to the right on small screens. Perhaps #80 would change the appearance in such a way that it wouldn't cause everything to do that?

Screen Shot 2019-12-11 at 09 49 41

rkivisto: Good point, I've reverted that change now @dwnusbaum

PR #59, opened 2019-01-25
Support retry in both fetch and doRetrieve

odavid:
Extending the retry logic to handle also cases where `scm.fetch(version, listener)` fails.
Got this Exception from time to time on our environment due to network issues. This PR handles such cases.

```
> git ls-remote -h git@OUR_HOST/global-pipeline-utils.git # timeout=10
hudson.plugins.git.GitException: Command "git ls-remote -h git@OUR_HOST/global-pipeline-utils.git" returned status code 128:
stdout: 
stderr: ssh_exchange_identification: read: Connection reset by peer
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:2042)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:1761)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:1666)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandWithCredentials(CliGitAPIImpl.java:1657)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.getRemoteReferences(CliGitAPIImpl.java:2877)
	at jenkins.plugins.git.AbstractGitSCMSource.retrieve(AbstractGitSCMSource.java:806)
	at jenkins.scm.api.SCMSource.fetch(SCMSource.java:617)
	at org.jenkinsci.plugins.workflow.libs.SCMSourceRetriever.retrieve(SCMSourceRetriever.java:82)
	at org.jenkinsci.plugins.workflow.libs.LibraryAdder.retrieve(LibraryAdder.java:157)
	at org.jenkinsci.plugins.workflow.libs.LibraryAdder.add(LibraryAdder.java:138)
	at org.jenkinsci.plugins.workflow.libs.LibraryDecorator$1.call(LibraryDecorator.java:125)
	at org.codehaus.groovy.control.CompilationUnit.applyToPrimaryClassNodes(CompilationUnit.java:1065)
	at org.codehaus.groovy.control.CompilationUnit.doPhaseOperation(CompilationUnit.java:603)
	at org.codehaus.groovy.control.CompilationUnit.processPhaseOperations(CompilationUnit.java:581)
[...]

PR #63, opened 2019-02-17
Handling RuntimeException during retry

odavid:
This PR handles cases that scm client fails to retrieve and the plugin throws unchecked exception.
In my case it was: `hudson.plugins.git.GitException`

```
 > git ls-remote -h git@XXX/CICD/global-pipeline-utils.git # timeout=10
hudson.plugins.git.GitException: Command "git ls-remote -h git@XXX/CICD/global-pipeline-utils.git" returned status code 128:
stdout: 
stderr: ssh_exchange_identification: Connection closed by remote host
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Relates to #59.


odavid:
@dwnusbaum - will appreciate if this can be approved as well.

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
```diff
@@ -108,7 +108,8 @@ public SCMSource getScm() {
             catch (InterruptedIOException e) {
                 throw e;
             }
-            catch (IOException e) {
+            // RuntimeException is a generalize case of untyped exceptions such as hudson.plugins.git.GitException
+            catch (RuntimeException | IOException e) {
```
dwnusbaum: At this point I would just catch all subclasses of `Exception`:
```suggestion
            catch (Exception e) {
```

dwnusbaum:
If we make this change here, should we add a matching change in workflow-cps? See [CpsScmFlowDefinition](https://github.com/jenkinsci/workflow-cps-plugin/blob/0fd80c496a8fd9181e08bb4ed5748ed6a9928904/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java#L154).

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
odavid: @dwnusbaum - you are right (: 
Fixed

odavid:
> If we make this change here, should we add a matching change in workflow-cps? See [CpsScmFlowDefinition](https://github.com/jenkinsci/workflow-cps-plugin/blob/0fd80c496a8fd9181e08bb4ed5748ed6a9928904/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScmFlowDefinition.java#L154).

I think we should, however, there, I can use retry logic within the pipeline. But when using global library in an implicit way, I cannot control it from within the pipeline.
Will create a PR there as well.

odavid:
@dwnusbaum - Will be thankful if you can approve that.
I also created https://github.com/jenkinsci/workflow-cps-plugin/pull/274, so it will be great if both will be approved...

dwnusbaum:
Thanks @odavid!

PR #64, opened 2019-03-07
Load steps from vars subfolders 

gerardsegarra:
Add functionality to load steps from vars's subfolders. This way you can classify the scripts in a more ordered way.


jglick:
I do not think this is a good idea. If you want structure, either use JVM packages in `src/`, or define subobjects inside variables.

PR #65, opened 2019-04-09
CustomDescribableModel can address SCMSourceRetriever.scm.id problem

jglick:
Downstream of https://github.com/jenkinsci/structs-plugin/pull/43.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,7 +125,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.18-rc265.27f956fa74ee</version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/43 -->
+            <version>1.18</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.28</version>
+        <version>3.42</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -64,10 +64,11 @@
     <properties>
         <revision>2.14</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.107.3</jenkins.version>
+        <jenkins.version>2.138.4</jenkins.version>
         <java.level>8</java.level>
+        <useBeta>true</useBeta>
         <workflow-cps-plugin.version>2.57</workflow-cps-plugin.version>
-        <git-plugin.version>3.6.4</git-plugin.version>
+        <git-plugin.version>4.0.0-rc</git-plugin.version>
         <scm-api-plugin.version>2.2.7</scm-api-plugin.version>
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
         <workflow-multibranch-plugin.version>2.10</workflow-multibranch-plugin.version>
@@ -117,7 +118,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.15</version>
+            <version>1.18-SNAPSHOT</version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/43 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.43</version>
+        <version>3.45</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -96,11 +96,6 @@
                     <artifactId>httpclient</artifactId>
                 </exclusion>
             </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>git-client</artifactId>
-            <version>3.0.0-rc</version> <!-- do not pick up an obsolete version from git-server -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -276,6 +271,15 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
+                <artifactId>git-client</artifactId>
+                <version>3.0.0-rc</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
     <build>
         <plugins>
             <plugin>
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
```diff
@@ -198,6 +202,19 @@ private static String getFilePathSuffix() {
             return descriptors;
         }
 
+        @Override public UninstantiatedDescribable customUninstantiate(UninstantiatedDescribable ud) {
+            Object scm = ud.getArguments().get("scm");
+            if (scm instanceof UninstantiatedDescribable) {
+                UninstantiatedDescribable scmUd = (UninstantiatedDescribable) scm;
+                Map<String, Object> scmArguments = new HashMap<>(scmUd.getArguments());
+                scmArguments.remove("id");
```
dwnusbaum: So IIUC, the problem is that we don't want this ID to be part of the model, since it should be unique to each instance?

jglick: Right. `SCMSource` does a weird trick that does not work like your typical `Describable`: it generates a random `id` each time you instantiate it, but also lets the ID be customized.

dwnusbaum:
Rebuilding because the Windows agent had connectivity issues.

pom.xml
```diff
@@ -64,11 +64,12 @@
...
         <revision>2.14</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.138.4</jenkins.version>
         <java.level>8</java.level>
+        <useBeta>true</useBeta>
         <workflow-cps-plugin.version>2.69</workflow-cps-plugin.version>
-        <git-plugin.version>3.6.4</git-plugin.version>
-        <scm-api-plugin.version>2.2.7</scm-api-plugin.version>
+        <git-plugin.version>4.0.0-rc</git-plugin.version>
```
dwnusbaum: PR looks good, but I want to wait for a stable release of `git` 4.x before merging the change (Commenting so I don't forget about this later). What would happen if we merged this as-is and released to the non-experimental update center? Broken instances for everyone because they can't satisfy the `git` dependency?

jglick: `git` is only a `test`-scoped dep, but `git-client` is a problem because it is in `dependencies` as `compile` scope:

```
Plugin-Dependencies: workflow-cps:2.69,workflow-scm-step:2.4,workflow-step-api:2.19,workflow-support:3.3,cloudbees-folder:6.1.0,credentials:2.1.14,git-client:3.0.0-rc,git-server:1.7,scm-api:2.3.0,script-security:1.58,structs:1.18
```

I suspect it is only there to satisfy `RequireUpperBoundsDeps`, in which case it would better be moved to `dependencyManagement` and then would not appear in the release at all. Would that suffice from your PoV?

dwnusbaum: Oops, I totally missed `git` being test-scope. Yeah if moving `git-client` into `dependencyManagement` gets it out of compile scope but tests still pass then I'd be fine to merge the change at that point.

jglick: OK, will try that.

dwnusbaum:
[`GrapeTest.var`](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/e7363ca89712f22a7e7c2ae55b919d6c40712a6f/src/test/java/org/jenkinsci/plugins/workflow/cps/global/GrapeTest.java#L80) failed in the CI build because of a script-security issue, not sure offhand how/if it relates to the changes in this PR:

> Scripts not permitted to use method org.apache.commons.math3.FieldElement add java.lang.Object. Administrators can decide whether to approve or reject this signature.

jglick:
Reproducible. Not dealing with it here, but someone is going to need to bisect [this stuff](https://github.com/jenkinsci/plugin-pom/compare/plugin-3.43...plugin-3.45) to find out why.

dwnusbaum:
> someone is going to need to bisect this stuff to find out why.

The problem is the jenkins-test-harness 2.50 update. Using plugin-pom 3.45 with jenkins-test-harness 2.49 the test passes. Looks like the issue is that https://github.com/jenkinsci/jenkins-test-harness/pull/135 added a dependency on jmh-core, which has a transitive dependency on commons-math3, which is the library grabbed in the test, so it breaks things (because it comes from the wrong class loader?). I guess the simplest fix would be to make the test `@Grab` some other library that is not already on the classpath somewhere to make the test work again.

`mvn dependency:tree` for `jenkins-test-harness` `master`:
```
[INFO] +- org.openjdk.jmh:jmh-core:jar:1.21:compile
[INFO] |  +- net.sf.jopt-simple:jopt-simple:jar:4.6:compile
[INFO] |  \- org.apache.commons:commons-math3:jar:3.2:compile

File: pom.xml [hunk 3]
```diff
@@ -276,6 +271,15 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.jenkins-ci.plugins</groupId>
```
basil: > Lots of other deps could probably be moved to dependencyManagement too; would be useful to have a tool which does this automatically based on source/bytecode scans.

I realize that the Jenkins community is fairly Maven-centric, but FYI Netflix's [Gradle Lint Plugin](https://github.com/nebula-plugins/gradle-lint-plugin) has this feature (c.f. [Unused Dependency Rule](https://github.com/nebula-plugins/gradle-lint-plugin/wiki/Unused-Dependency-Rule)).

jglick: I think it is likely there is a Maven equivalent, I just have not looked for it yet.

jglick:
Note that as described in JIRA, this will _not_ magically fix things for `configuration-as-code`, which unfortunately does not use `DescribableModel` but reimplements a similar reflection system.

pom.xml
basil: > I think it is likely there is a Maven equivalent, I just have not looked for it yet.

Looks like the [Maven Dependency Plugin](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html) can list unused declared dependencies, though (just like the Gradle version) there seem to be too many false positives for it to be usable as part of any kind of automated enforcement.

jglick: The problem of false positives is compounded by the heavy use of reflection, Groovy views, etc. etc. in Jenkins sources.

pom.xml
dwnusbaum: Ok, merged this a bit prematurely because I did not actually run `mvn dependency:tree`. This library has a compile-scope dependency on git-server which itself has a compile-scope dependency on git-client, so moving `git-client` to `dependencyManagement` does not fix the problem, and so the plugin currently has an experimental release as a compile-scope dependency. I'm not sure to what extent `git-server` uses `git-client` or if there is an easy fix for this. Maybe safest to revert the merge for now?

jglick: > moving `git-client` to `dependencyManagement` does not fix the problem

It does because it does not have a _direct_ dependency on `git-client`, so it does not appear in the manifest.

```sh
mvn -DskipTests clean package
unzip -p target/workflow-cps-global-lib.hpi META-INF/MANIFEST.MF | perl -p -0777 -e 's/\r?\n //g' | fgrep Plugin-Dependencies
```

⇒

```
Plugin-Dependencies: workflow-cps:2.69,workflow-scm-step:2.4,workflow-step-api:2.19,workflow-support:3.3,cloudbees-folder:6.1.0,credentials:2.1.14,git-server:1.7,scm-api:2.3.0,script-security:1.58,structs:1.18
```

dwnusbaum: Ok makes sense, in an actual instance the version of git-client required by the plugin will be whatever git-server 1.7 requires because of how Jenkins resolves dependencies.

jglick: Rght.

PR #68, opened 2019-04-24
[JENKINS-43802] No context available for credentials from SCMSource.retrieve(String, TaskListener)

jglick:
[JENKINS-43802](https://issues.jenkins-ci.org/browse/JENKINS-43802)

Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/65.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <java.level>8</java.level>
         <workflow-cps-plugin.version>2.67</workflow-cps-plugin.version>
         <git-plugin.version>3.6.4</git-plugin.version>
-        <scm-api-plugin.version>2.5.0-SNAPSHOT</scm-api-plugin.version> <!-- TODO https://github.com/jenkinsci/scm-api-plugin/pull/65 -->
+        <scm-api-plugin.version>2.5.0-rc464.915933672136</scm-api-plugin.version> <!-- TODO https://github.com/jenkinsci/scm-api-plugin/pull/65 -->
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
         <workflow-multibranch-plugin.version>2.10</workflow-multibranch-plugin.version>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -68,7 +68,7 @@
         <java.level>8</java.level>
         <workflow-cps-plugin.version>2.67</workflow-cps-plugin.version>
         <git-plugin.version>3.6.4</git-plugin.version>
-        <scm-api-plugin.version>2.2.7</scm-api-plugin.version>
+        <scm-api-plugin.version>2.5.0-rc464.915933672136</scm-api-plugin.version> <!-- TODO https://github.com/jenkinsci/scm-api-plugin/pull/65 -->
```
bitwiseman:
```suggestion
        <scm-api-plugin.version>2.6.2</scm-api-plugin.version>
```

jglick:
Will not work because of the gratuitously high core dependency from https://github.com/jenkinsci/scm-api-plugin/pull/61.

dwnusbaum:
Yeah, we could update here, but given that 2.164.3 is only two months old I'd prefer to bump scm-api back down to 2.107.x or whatever it actually needed (or at least a version like 2.138.4 that has been out for a bit longer).

pom.xml
```diff
@@ -69,7 +69,7 @@
         <useBeta>true</useBeta>
         <workflow-cps-plugin.version>2.71</workflow-cps-plugin.version>
         <git-plugin.version>4.0.0-rc</git-plugin.version>
-        <scm-api-plugin.version>2.4.0</scm-api-plugin.version>
+        <scm-api-plugin.version>2.6.2</scm-api-plugin.version>
```
bitwiseman:
```suggestion
        <scm-api-plugin.version>2.6.3</scm-api-plugin.version>
```

dwnusbaum:
Rebuilding due to Windows agent failure.

PR #72, opened 2019-06-24
[JENKINS-43802] Fix retrieval to use proxy owner so credentials get resolved

stephenc:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/70

I have not addressed the dependency bound enforcer issues


jglick:
FTR, this was proposed as an alternative to #68.

jglick:
Upstream was closed (CC @bitwiseman).

PR #73, opened 2019-06-25
Make GrapeTest.var work with new versions of jenkins-test-harness

dwnusbaum:
In #65, we discovered that `GrapeTest.var` breaks with new versions of jenkins-test-harness. See https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/65#issuecomment-505091612 for details. This PR fixes the issue by switching from `commons-math3` to `commons-primitives`, which was already being used in other tests here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.43</version>
+        <version>3.46</version>
         <relativePath />
     </parent>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.43</version>
+        <version>3.46</version>
```
jglick:
```suggestion
        <version>3.47</version>
```

PR #71, opened 2019-06-04
[JENKINS-56682] Reproduce and fix issue using types from shared libraries in Script initializers

dwnusbaum:
Same underlying problem as [JENKINS-56682](https://issues.jenkins-ci.org/browse/JENKINS-56682). See https://github.com/jenkinsci/script-security-plugin/pull/259 and https://github.com/jenkinsci/workflow-cps-plugin/pull/297.

This PR updates a bunch of dependencies since I had already had a wip commit to do that locally. More importantly, it adds a new test whose failing stack trace looks like: 

```
org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use new pkg.Foo
	at org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.StaticWhitelist.rejectNew(StaticWhitelist.java:271)
	at org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onNewInstance(SandboxInterceptor.java:174)
	at org.kohsuke.groovy.sandbox.impl.Checker$3.call(Checker.java:198)
	at org.kohsuke.groovy.sandbox.impl.Checker.checkedConstructor(Checker.java:203)
	at org.kohsuke.groovy.sandbox.impl.Checker$checkedConstructor.callStatic(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallStatic(CallSiteArray.java:56)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:194)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:214)
	at WorkflowScript.<init>(WorkflowScript:3)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at org.codehaus.groovy.runtime.InvokerHelper.createScript(InvokerHelper.java:434)
Caused: groovy.lang.GroovyRuntimeException: Failed to create Script instance for class: class WorkflowScript. Reason
[...]

PR #75, opened 2019-07-31
[JENKINS-58771] Do not bundle JARs from apache-httpcomponents-client-4-api

jglick:
Amends #71. Noticed via https://github.com/jenkinsci/maven-hpi-plugin/pull/130. Without this, `apache-httpcomponents-client-4-api` is in `compile` scope (despite being only here due to `workflow-basic-steps`, a `test`-scoped dep), and thus various things like `http-core` are as well, so `hpi:hpi` gets confused (since the direct dep `workflow-basic-steps` is not in its listed of recognized plugin dependencies) and bundles them, bloating the archive.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,8 +96,8 @@
             <artifactId>git-server</artifactId>
             <version>1.7</version>
             <exclusions>
-                 <exclusion>
-                 <!-- Workaround for packaging bug in maven-hpi-plugin, see discussion in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/75 -->
+                <exclusion>
+                    <!-- TODO JENKINS-58771: workaround for bug in hpi:hpi -->
                     <groupId>org.jenkins-ci.plugins</groupId>
                     <artifactId>apache-httpcomponents-client-4-api</artifactId>
                 </exclusion>
--- a/pom.xml
+++ b/pom.xml
@@ -97,8 +97,8 @@
             <version>1.7</version>
             <exclusions>
                 <exclusion>
-                    <groupId>org.apache.httpcomponents</groupId>
-                    <artifactId>httpclient</artifactId>
+                    <groupId>org.jenkins-ci.plugins</groupId>
+                    <artifactId>apache-httpcomponents-client-4-api</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
@@ -271,12 +271,6 @@
     </dependencies>
     <dependencyManagement>
         <dependencies>
-            <dependency>
-                <groupId>org.jenkins-ci.plugins</groupId>
-                <artifactId>apache-httpcomponents-client-4-api</artifactId>
-                <version>4.5.5-3.0</version>
-                <scope>test</scope>
-            </dependency>
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>git-client</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -275,6 +275,7 @@
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>apache-httpcomponents-client-4-api</artifactId>
                 <version>4.5.5-3.0</version>
+                <scope>test</scope>
             </dependency>
             <dependency>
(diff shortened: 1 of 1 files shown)


jglick:
I think what was happening was that there was a second-level dep on `apache-httpcomponents-4-api` via `workflow-basic-steps` (`test`) and a third-level dep via `git-server` → `git-client` (`compile`). Since at least one trail was `compile`-scoped, Maven treated it as a `compile`-scoped dependency. Yet `maven-hpi-plugin` only exclude JARs from bundling when the second artifact in the dependency trail (i.e., the direct dependency of “this” plugin) is a plugin, and it only checks `compile`-scoped artifacts for pluginhood. That smells like a bug in `AbstractHpiMojo`, though not one previously reported (e.g., in JENKINS-53957) to my knowledge. Fixing that would perhaps be the better fix, though it is riskier and would require an IT.

dwnusbaum:
> Yet maven-hpi-plugin only exclude JARs from bundling when the second artifact in the dependency trail (i.e., the direct dependency of “this” plugin) is a plugin, and it only checks compile-scoped artifacts for pluginhood

I am confused. `workflow-cps-global-lib`, `git-server`, `git-client`, and `apache-httpcomponents-client-4-api` are all plugins and depend on each other in `compile` scope in that order. That seems like it should work fine by itself with no exclusions.

Are you saying that because the path to `apache-httpcomponents-client-4-api` through `workflow-basic-test` is `test` scope and is "closer" than the `compile` scope path through `git-server`, `maven-hpi-plugin` doesn't consider excluding `workflow-basic-test`'s dependencies from being packaged, and then because `apache-httpcomponents-client-4-api` is actually a `compile` scope dependency it ends up being packaged?

Any idea if other plugins have similar issues as well? This doesn't seem like a particularly unusual case.

pom.xml
```diff
@@ -271,11 +271,6 @@
     </dependencies>
     <dependencyManagement>
         <dependencies>
-            <dependency>
-                <groupId>org.jenkins-ci.plugins</groupId>
-                <artifactId>apache-httpcomponents-client-4-api</artifactId>
```
dwnusbaum: So, IIUC, if not for the bug in `AbstractHpiMojo `, this would have been correct? And the exclusion for  `org.apache.httpcomponents:httpclient` from `git-server` was due to an outdated `git-client` dep at some point and could be deleted if not for the bug? Or are `exclusions` in a POM specific only to that POM and not to upstream POMs, so we have to reapply them here?

The exclusion of `apache-httpcomponents-client-4-api` doesn't matter at runtime because it's actually going to come in as a plugin through the `git-server` to `git-client` dependency chain, right? This is just a hack to fix the packaging here?

jglick:
> Are you saying that…

Yes.

> Any idea if other plugins have similar issues as well?

No clue. We do not routinely check for bogus bundled JARs, only by accident.

pom.xml
jglick: If the mojo were fixed, I _think_ this could have been kept and the `exclusion` removed.

Exclusions are heritable.

jglick: > doesn't matter at runtime

Right.

pom.xml
```diff
@@ -97,8 +97,8 @@
             <version>1.7</version>
             <exclusions>
                 <exclusion>
```
dwnusbaum: Can we add a comment like the following?

```suggestion
                 <exclusion>
                 <!-- Workaround for packaging bug in maven-hpi-plugin, see discussion in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/75 -->
```

jglick:
Should recheck after https://github.com/jenkinsci/maven-hpi-plugin/pull/140.

jglick:
…and thence https://github.com/jenkinsci/plugin-pom/pull/234

pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.47</version>
+        <version>3.50</version>
```
jglick: Thanks @Vlatombe!

PR #76, opened 2019-08-14
adding the ability for users to override implicitly loaded global libraries.

willcrain1:
In my company we use this shared pipeline library plugin, and I find it hard to test changes to shared pipeline libraries unless I have write access to their repository and am able to create a branch, then I can refer to the changes with @Library('libraryname@branchname').

Currently it seems there's no way to import a fork of a library to override an implicitly loaded shared library.  I'd use this functionality frequently to test any changes made in our shared libraries.

I was able to test this PR by building hpi files and importing them into the jenkinsci official docker container.  I then created a shared library here:  https://github.com/willcrain1/test-pipeline-library and forked it here as well and made minor changes:  https://github.com/jenkinspipelinetesting/test-pipeline-library

I executed the libraries in this Jenkinsfile: https://github.com/willcrain1/test-pipeline/blob/master/Jenkinsfile

@jglick @kohsuke any feedback is appreciated.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
@@ -89,6 +89,7 @@ public class LibraryStep extends AbstractStepImpl {
     private final String identifier;
     private Boolean changelog = true;
     private LibraryRetriever retriever;
+    private Boolean existingLibrariesUsed = true;
 
     @DataBoundConstructor public LibraryStep(String identifier) {
         this.identifier = identifier;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
```diff
@@ -105,6 +106,15 @@ public LibraryRetriever getRetriever() {
     public Boolean getChangelog() {
         return changelog;
     }
+
+    public Boolean getExistingLibrariesUsed() {
```
jglick: Getters and setters should take `boolean`, not `Boolean`.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
```diff
@@ -89,6 +89,7 @@
     private final String identifier;
     private Boolean changelog = true;
     private LibraryRetriever retriever;
+    private Boolean existingLibrariesUsed = true;
```
jglick: This seems like a risky default.

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b PLEASE-SELECT-BRANCH-NAME
git pull https://github.com/willcrain1/workflow-cps-global-lib-plugin master
```

PR #77, opened 2019-08-16
Adding LICENSE file.

kshithijiyer:


dwnusbaum:
I'm not sure about this. In pom.xml we already declare that we are using the MIT license: https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/ff614b7aab5d5ec333fc74c6d62290958de62912/pom.xml#L42

Is that not enough? Each file should have its own license header with the author who submitted the change. If we created a file in the root, would we need to include all of those authors in the main file and then update the date?

kshithijiyer:
It is good to have a dedicated License file. Not every project will have a pom.xml and when someone is looking for a place to find what license a project is open sourced under than he'll generally look for a license file.

dwnusbaum:
@oleg-nenashev Do you know though if we this is the right way to do it, given there are many independent copyright owners listed in the headers in this plugin? I am not sure if it matters or not:

> If we created a file in the root, would we need to include all of those authors in the main file and then update the date?

oleg-nenashev:
I just append "and all other contributors" in the end of the line.

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 CloudBees, Inc.
```
oleg-nenashev:
```suggestion
Copyright (c) 2016-2019 CloudBees, Inc. and all other contributors
```

PR #83, opened 2019-09-10
Support maven folder layout for src

v1v:
## Highlights
- Use the [standard folder layout](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) to support UTs, ITs and other stages out of the box.
- Even though the `src/main/groovy` is not within the standard layout, I'd say it seems it's the de facto standard.
- Backward compatible.

## Questions
- Not sure what other UTs should be added for this particular feature.

## Tasks
- [ ] Update docs https://github.com/jenkins-infra/jenkins.io/blob/master/content/doc/book/pipeline/shared-libraries.adoc#directory-structure

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GroovyShellDecoratorImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/cps/global/GroovyShellDecoratorImpl.java
@@ -29,6 +29,7 @@ public class GroovyShellDecoratorImpl extends GroovyShellDecorator {
                 try {
                     shell.getClassLoader().addURL(new File(repo.workspace,"src").toURI().toURL());
                     shell.getClassLoader().addURL(new File(repo.workspace, UserDefinedGlobalVariable.PREFIX).toURI().toURL());
+                    shell.getClassLoader().addURL(new File(repo.workspace,"src/main/groovy").toURI().toURL());
                 } catch (MalformedURLException e) {
                     throw new AssertionError(e);
                 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -90,7 +90,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
             // Resuming a build, so just look up what we loaded before.
             for (LibraryRecord record : action.getLibraries()) {
                 FilePath libDir = new FilePath(execution.getOwner().getRootDir()).child("libs/" + record.name);
-                for (String root : new String[] {"src", "vars"}) {
+                for (String root : new String[] {"src", "src/main/groovy", "vars"}) {
                     FilePath dir = libDir.child(root);
                     if (dir.isDirectory()) {
                         additions.add(new Addition(dir.toURI().toURL(), record.trusted));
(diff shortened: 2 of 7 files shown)


v1v:
@jenkinsci/code-reviewers , what do you think?

v1v:
> IUC, anything under src/test would still be available in Jenkins because we add the URL for src, right (which could be problematic if your test code is not intended to be used)? Or no, because the packages would be wrong, so if you tried to use them things would break?

Just to clarify this PR doesn't add the `src` folder as that's already the current design. This PR does add the `src/main/groovy` subfolder to the classpath. So therefore, no changes in how the  `src/test` might be behaving. Likely, the description/title might be a bit misleading what I've done in this PR.

dwnusbaum:
> So therefore, no changes in how the src/test might be behaving

Yes, my point is more that if `src/main/groovy` is supported, users might start using `src/test` where they previously would have used `test/`, making test code available accidentally because `src` is available.

v1v:
That's indeed the current use case and supported already:
- https://github.com/jenkins-infra/pipeline-library/tree/master/src/test/groovy

The JenkinsPipelineUnit project provides a similar folder layout:
- https://github.com/jenkinsci/JenkinsPipelineUnit/blob/master/README.md#configuration

I have not faced any issues in any of the projects that I've been working with when using the JenkinsUnitTests and the shared library with the folder `src/test`.

v1v:
@dwnusbaum, not sure whether my previous comment addressed your concerns, please let me know if you need further feedback.

Other than that, is this PR ready to be merged? 🙏

v1v:
@dwnusbaum , let me know if you need further clarification from my comments.

Thanks

src/main/java/org/jenkinsci/plugins/workflow/cps/global/GroovyShellDecoratorImpl.java
```diff
@@ -29,6 +29,7 @@ public void configureShell(CpsFlowExecution context, GroovyShell shell) {
                 try {
                     shell.getClassLoader().addURL(new File(repo.workspace,"src").toURI().toURL());
                     shell.getClassLoader().addURL(new File(repo.workspace, UserDefinedGlobalVariable.PREFIX).toURI().toURL());
+                    shell.getClassLoader().addURL(new File(repo.workspace,"src/main/groovy").toURI().toURL());
```
dwnusbaum: Because the `src` path includes this path, I was curious about what would happen if you imported a class in `src/main/groovy` twice, once using the class, and once using `main.groovy` as the package for the class. It looks like it fails with a `LinkageError`, which is safe, so I think that's fine, although a bit weird.

Here is a test case that fails with `java.lang.LinkageError: loader (instance of  org/jenkinsci/plugins/workflow/cps/CpsGroovyShell$CleanGroovyClassLoader): attempted  duplicate class definition for name: "Foo"` demonstrating what I mean:

```
@Test public void srcClassViaMultiplePaths() throws Exception {
    sampleRepo.init();
    sampleRepo.write("src/main/groovy/Foo.groovy", "class Foo { }");
    sampleRepo.git("add", "src");
    sampleRepo.git("commit", "--message=init");
    GlobalLibraries.get().setLibraries(Collections.singletonList(
        new LibraryConfiguration("lib",
            new SCMSourceRetriever(new GitSCMSource(null, sampleRepo.toString(), "", "*", "", true)))));
    WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
    p.setDefinition(new CpsFlowDefinition(
            "@Library('lib@master')\n +
            "import Foo\n" +
            "new main.groovy.Foo()", true));
    r.buildAndAssertSuccess(p);
}
```

timja:
I would like this change, it means less config to make IDEs work for running tests, (not that I’ve ever managed to make it work fully for shared libraries)

pault1337:
I agree with @timja. Having this change allows to fully take advantages of your IDE. On Intellij for e.g. you'll have code completion, navigation between classes, unit test generation and run, etc., right away. It's very comfortable to have this (we are actually running the plugin with this change).

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b feature/support-maven-layout
git pull https://github.com/v1v/workflow-cps-global-lib-plugin feature/support-maven-layout
```

PR #87, opened 2019-11-21
[JENKINS-60245] Potential fix for @Library annotation in load()'d scripts

awittha:
This is a potential fix for the issue described in [JENKINS-60245](https://issues.jenkins-ci.org/browse/JENKINS-60245). I call it a "potential" fix because while it seems to fix all the broken use-cases and cause no regressions, I am not sure if there is a more architecturally-sound means to accomplish the same ends.

Please see the JIRA issue for a complete description of the problem.

These changes add a `String scope` to `LibrariesAction`, and attach one action to a build _per loaded script_.

When recording shared libraries that were mentioned in a `@Library` annotation, the script being processed is used as the scope of the `LibrariesAction`. This enables the plugin to know that it should index the contents of `@Library` annotations in newly-loaded scripts.

When processing libraries to actually add their resources to the classpath, every `LibrariesAction` for the build is iterated over.

This allows scripts loaded with the `load(...)` step to add shared libraries with the `@Library(...)` annotation.

I had to _update_ some existing unit tests, but did not write a regression test for the new behavior. If this implementation of a solution to JENKINS-60245 is determined to be generally-sound & on-track for merging into the plugin, I will write such a test.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/ClasspathAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/ClasspathAdder.java
@@ -27,9 +27,10 @@ package org.jenkinsci.plugins.workflow.libs;
 import groovy.lang.GroovyShell;
 import hudson.ExtensionPoint;
 import java.net.URL;
-import java.util.List;
 import java.util.HashMap;
+import java.util.List;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowExecution;
 
 /**
@@ -58,12 +59,20 @@ public abstract class ClasspathAdder implements ExtensionPoint {
     }
 
     /**
+     * @see #add(String, CpsFlowExecution, List, HashMap)
+     */
+    public @Nonnull List<Addition> add(@Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception {
+        return add( null, execution, libraries, changelogs );
+    }
+    
+    /**
      * May add to the classpath.
+     * @param scope identifier for the source file that the addition came from
      * @param execution a running build (possibly newly started, possibly resumed)
      * @param libraries aggregated entries from all encountered {@link Library#value} (will be empty if {@link Library} is never used at all); an implementation should remove entries it “claims”
      * @return a possibly empty list of additions
      * @throws Exception for whatever reason (will fail compilation)
      */
-    public abstract @Nonnull List<Addition> add(@Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception;
+    public abstract @Nonnull List<Addition> add(@Nullable String scope, @Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception;
 
 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
@@ -39,8 +39,15 @@ import java.util.List;
 class LibrariesAction extends InvisibleAction {
 
     private final List<LibraryRecord> libraries;
-
+    
+    private final String scope;
+    
     LibrariesAction(List<LibraryRecord> libraries) {
+        this(null, libraries);
+    }
+
+    LibrariesAction(String scope, List<LibraryRecord> libraries) {
+        this.scope = scope;
         this.libraries = libraries;
     }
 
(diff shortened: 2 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/ClasspathAdder.java
```diff
@@ -57,13 +58,21 @@ void addTo(@Nonnull CpsFlowExecution execution) {
...
     /**
      * May add to the classpath.
+     * @param scope identifier for the source file that the addition came from
      * @param execution a running build (possibly newly started, possibly resumed)
      * @param libraries aggregated entries from all encountered {@link Library#value} (will be empty if {@link Library} is never used at all); an implementation should remove entries it “claims”
      * @return a possibly empty list of additions
      * @throws Exception for whatever reason (will fail compilation)
      */
-    public abstract @Nonnull List<Addition> add(@Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception;
+    public abstract @Nonnull List<Addition> add(@Nullable String scope, @Nonnull CpsFlowExecution execution, @Nonnull List<String> libraries, @Nonnull HashMap<String, Boolean> changelogs) throws Exception;
```
awittha: New `add(...)` method added to the `ClasspathAdder` interface, to allow for scoping the source of an addition.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
```diff
@@ -39,8 +39,15 @@
...
-
+    
+    private final String scope;
+    
     LibrariesAction(List<LibraryRecord> libraries) {
+        this(null, libraries);
+    }
+
+    LibrariesAction(String scope, List<LibraryRecord> libraries) {
+        this.scope = scope;
```
awittha: New constructor & field added to `LibrariesAction` to allow recording of "scope" - what source file the libraries in this LibrariesAction are from.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -85,7 +85,19 @@
             librariesUnparsed.put(parsed[0], library);
         }
         List<Addition> additions = new ArrayList<>();
-        LibrariesAction action = build.getAction(LibrariesAction.class);
+        LibrariesAction action = null;
+        
+        for( LibrariesAction laction : build.getActions( LibrariesAction.class ) ) {
+            if( laction.getScope() != null && laction.getScope().equals( scope ) ) {
```
awittha: Previously, as there was only ever one `LibrariesAction` per build, if it was present, it was concluded that this was a resumed build and that action could be used.

Now, there may be multiple actions so we must find the one whose `scope` equals the `scope` passed into this `add` invocation.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -85,7 +85,19 @@
...
+                // found the action that's absolutely right for this scope
+                action = laction;
+                break;
+            } else if( laction.getScope() == null ) {
+                // generic LibrariesAction may be overridden later if there's a perfect match.
+                action = laction;
```
awittha: It is possible that a build was suspended before these code changes landed, in which case the "right" LibrariesAction will have a `null` scope.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -131,12 +143,14 @@
...
         }
         // Record libraries we plan to load. We need LibrariesAction there first so variables can be interpolated.
-        build.addAction(new LibrariesAction(new ArrayList<>(librariesAdded.values())));
-        // Now actually try to retrieve the libraries.
-        for (LibraryRecord record : librariesAdded.values()) {
-            listener.getLogger().println("Loading library " + record.name + "@" + record.version);
-            for (URL u : retrieve(record.name, record.version, retrievers.get(record.name), record.trusted, record.changelog, listener, build, execution, record.variables)) {
-                additions.add(new Addition(u, record.trusted));
+        if( librariesAdded.size() > 0 ) {
+            build.addAction(new LibrariesAction(new ArrayList<>(librariesAdded.values())));
```
awittha: Avoid adding a `LibrariesAction` with 0 library definitions in it.

dwnusbaum:
Thanks for the PR! I'm not totally sure about this, here are a few of the immediate questions that came to mind:

* Would this also work for a shared library that references another shared library via `@Library`, or is this only for scripts loaded via the `load` step?
* With this change, if there is more than one `load` step, and each loaded script loads the same shared library, would we have multiple instances of the shared library loaded in the Pipeline?
* In the use case you give in [JENKINS-60245](https://issues.jenkins-ci.org/browse/JENKINS-60245), why not just drop the `load` step entirely and have your `Jenkinsfile` reference the library? I am guessing you simplified your actual use case?

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b JENKINS-60245
git pull https://github.com/awittha/workflow-cps-global-lib-plugin JENKINS-60245
```

PR #88, opened 2019-12-18
Bump plugin parent POM

basil:
Bumps the [plugin parent POM](https://github.com/jenkinsci/plugin-pom) from 3.50 to 3.54.

See [the changelog](https://github.com/jenkinsci/plugin-pom/releases) and the full diff in [the compare view](https://github.com/jenkinsci/plugin-pom/compare/plugin-3.50...plugin-3.54).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,6 +70,7 @@
         <workflow-cps-plugin.version>2.71</workflow-cps-plugin.version>
         <git-plugin.version>4.0.0-rc</git-plugin.version>
         <scm-api-plugin.version>2.6.3</scm-api-plugin.version>
+        <subversion-plugin.version>2.13.0</subversion-plugin.version>
         <workflow-scm-step-plugin.version>2.7</workflow-scm-step-plugin.version>
         <workflow-multibranch-plugin.version>2.21</workflow-multibranch-plugin.version>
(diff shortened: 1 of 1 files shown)


basil:
To fix `LibraryAdderTest.properSvn` I had to apply a similar fix as in jenkinsci/workflow-multibranch-plugin#92.

pom.xml
```diff
@@ -28,15 +28,15 @@
...
+        <version>3.54</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-cps-global-lib</artifactId>
     <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Shared Groovy Libraries</name>
-    <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+Shared+Groovy+Libraries+Plugin</url>
+    <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
```
jglick: This should be paired with a patch to `README.md` importing the wiki content using https://jenkins-wiki-exporter.jenkins.io/

basil: I have no interest in doing that, so I will just revert this change and let the plugin maintainer take care of it.

PR #93, opened 2020-04-09
Bump commons-lang3 version to 3.9

timja:
Sync it with the testharness-html-unit version,
Otherwise we get conflicts on bom

error when introducing 4.x parent pom in bom:
```
Require upper bound dependencies error for org.apache.commons:commons-lang3:3.7 paths to dependency are:
+-io.jenkins.tools.bom:sample:7-SNAPSHOT
  +-org.jenkins-ci.plugins.workflow:workflow-cps-global-lib:2.15
    +-org.apache.commons:commons-lang3:3.7
and
+-io.jenkins.tools.bom:sample:7-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-test-harness:2.62
    +-org.jenkins-ci.main:jenkins-test-harness-htmlunit:2.36.0-1
      +-org.apache.commons:commons-lang3:3.9
and
+-io.jenkins.tools.bom:sample:7-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-test-harness:2.62
    +-org.jenkins-ci.main:jenkins-test-harness-htmlunit:2.36.0-1
      +-org.apache.commons:commons-text:1.7
        +-org.apache.commons:commons-lang3:3.9
```

Very similar error here:
https://github.com/jenkinsci/bom/pull/206

Thoughts?
@jglick @dwnusbaum @bitwiseman

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.7</version>
+            <version>3.9</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>


timja:
> Sounds good to me, although I wonder much work it would be to replace the usages in `LibraryStep` with something else so we can just drop the dependency completely? It looks like it's used in the context of `GroovyObjectSupport`, so maybe we could use `org.codehaus.groovy.runtime.InvokerHelper.invokeConstructorOf` and `invokeStaticMethod` instead?

opened a PR to see if it will work: https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/94

dwnusbaum:
Closing since #94 was merged instead.

PR #94, opened 2020-04-09
Remove commons-lang3

timja:
See: https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/93#pullrequestreview-391129809

@jglick @dwnusbaum @bitwiseman

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
@@ -41,7 +41,6 @@ import hudson.security.AccessControlled;
 import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


timja:
Ran a basic smoke test which seemed to work, the library was loaded:
```
library 'pipeline'
buildPlugin()
```

dwnusbaum:
@timja Thanks for filing the PR! I think that normal cases should work fine, my main concern is if @jglick had some particular reason not to use `InvokerHelper`. Maybe it was not used so that Groovy-specific features would not work with the library step to avoid potential sandbox issues?

jglick:
> if @jglick had some particular reason not to use `InvokerHelper`

To the extent I can recall now, I think I just did not realize it was available.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryStep.java
```diff
@@ -265,16 +263,10 @@ public AutoCompletionCandidates doAutoCompleteIdentifier(@AncestorInPath ItemGro
         @Override public Object invokeMethod(String name, Object _args) {
             Class<?> c = loadClass(prefix + clazz);
             Object[] args = _args instanceof Object[] ? (Object[]) _args : new Object[] {_args}; // TODO why does Groovy not just pass an Object[] to begin with?!
```
dwnusbaum, line 265: nit: `InvokerHelper` handles this conversion internally via `InvokerHelper.asArray`, so I think this line could be removed now.

PR #100, opened 2020-09-25
[JENKINS-41037] Prevent ConcurrentModificationException serializing LibrariesAction.libraries

jglick:
Should solve the cases of [JENKINS-41037](https://issues.jenkins-ci.org/browse/JENKINS-41037) associated with the `libraries` step, which is most of them. Alternative to https://github.com/x-stream/xstream/pull/222. Not tested (do not know how to reproduce problem).

Note that `CopyOnWriteList` is an alternative fix, for cases where it would be awkward to make the collection immutable, but in this case there is only one piece of code that was ever mutating this list so it is most straightforward to fix that.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
@@ -31,6 +31,7 @@ import hudson.model.InvisibleAction;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


dwnusbaum (approved):
So the problem was that the `notifyListeners` call from the step prior to the `library` step was in the middle of saving the action (while it was saving the `WorkflowRun`) right as the `library` step started executing and modified the field? I am somewhat surprised that this occurred frequently, but the proposed fix looks fine, thanks!

jglick:
I am not sure offhand what was calling `WorkflowRun.save`; maybe varied.

PR #84, opened 2019-09-16
JENKINS-38609, Shared Library non ROOT support - initial implementation

yizhuan:
This is an implementation for JENKINS-38609, which makes it possible to load libraries from a specified relative path.

PR #101, opened 2020-10-19
[JENKINS-38609] Shared Library non ROOT support

dwnusbaum:
See [JENKINS-38609](https://issues.jenkins-ci.org/browse/JENKINS-38609). This is a refiled version of #99 from @heyleke with various minor changes, for example, `libBasePath` was renamed to `libraryPath`, the setters for `libraryPath` append `/` if needed, localization support was added, help text was added for the new field, minor changes were made to the regex that prevents `..` from being used in the library path, some formatting changes were reverted, and the UI for the SCM selection was changed.


src/main/resources/org/jenkinsci/plugins/workflow/libs/SCMRetriever/config.properties
```diff
@@ -0,0 +1,23 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+libraryPath=Library Path (optional)
```
dwnusbaum, line 23: Normally we would not need to add a separate file, but I thought it was useful to clarify that the field was optional and literal parentheses cannot be used in the `${%...}` syntax.

src/main/resources/org/jenkinsci/plugins/workflow/libs/SCMRetriever/config.jelly
```diff
@@ -25,5 +25,8 @@ THE SOFTWARE.
 
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
-    <f:descriptorRadioList varName="scm" instance="${instance.scm}" descriptors="${descriptor.SCMDescriptors}" title="${%Source Code Management}"/>
+    <f:dropdownDescriptorSelector field="scm" descriptors="${descriptor.SCMDescriptors}" title="${%Source Code Management}"/>
```
dwnusbaum, line 28: This change (same for `SCMSourceRetriever`) has a non-trivial effect on the layout (changes it from radio buttons to a dropdown), but I think we have to make this change or else the layout with the new field is really confusing. We should do some interactive testing to make sure everything still works correctly. I am not sure if there was a reason that radio buttons were used in the first place, maybe it was to reduce the level of nesting?

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
```diff
@@ -157,14 +190,19 @@ static void doRetrieve(String name, boolean changelog, @Nonnull SCM scm, FilePat
                 delegate.checkout(run, lease.path, listener, node.createLauncher(listener));
                 return null;
             });
-            // Cannot add WorkspaceActionImpl to private CpsFlowExecution.flowStartNodeActions; do we care?
-            // Copy sources with relevant files from the checkout:
+            if (libraryPath == null) {
+                libraryPath = ".";
+            } else if (PROHIBITED_DOUBLE_DOT.matcher(libraryPath).matches()) {
+                throw new AbortException("Library path may not contain '..'");
```
dwnusbaum, line 196: I intentionally did not use the localized message here, because I think we don't normally localize build log content (since it is not produced in a context where this an active user), but I do not remember for sure if this is the right thing to do.

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
```diff
@@ -74,9 +79,25 @@
 
     @SuppressFBWarnings(value="MS_SHOULD_BE_FINAL", justification="Non-final for write access via the Script Console")
     public static boolean INCLUDE_SRC_TEST_IN_LIBRARIES = Boolean.getBoolean(SCMSourceRetriever.class.getName() + ".INCLUDE_SRC_TEST_IN_LIBRARIES");
+    /**
+     * Matches ".." in positions where it would be treated as the parent directory.
+     *
+     * <p>Used to prevent {@link #libraryPath} from being used for directory traversal.
+     */
+    static final Pattern PROHIBITED_DOUBLE_DOT = Pattern.compile("(^|.*[\\\\/])\\.\\.($|[\\\\/].*)");
```
dwnusbaum, line 87: Slightly expanded from the original version to handle cases with only a leading or trailing slash or just `..`, since I think all of those could be problematic for our use case.

heyleke:
Looks good to me, when this gets approved, I have a further cleanup of this code ready where the duplication between SCMRetriever and SCMSourceRetriever is handled, but please don't hold this PR for that.

One of the tests seem to fail on timeout of 180s, is there a rule and strategy on how to cope with that?

heyleke (approved):

heyleke:
Is it normal in the Jenkins community that Pull Request from new contributors go rotten, this PR, which was orignially from another user, reworked by myself in #99 and then reworked by @dwnusbaum in #101 is now officially 1 month 'untouched'... I feel a bit lost how to get this merged, asking @oleg-nenashev for advice how to handle this in the Jenkins Community. @dwnusbaum has been great in guiding me, but now the work seems to get stuck?

dwnusbaum:
@heyleke Remember that if something is critical for you, you can always build a version of the plugin based on the PR and run that on your Jenkins instance. I am sorry this has been delayed, but I have extremely limited time to look into anything other than bug fixes or regressions for OSS Jenkins plugins these days. When I filed this PR I thought I would have some time to get it finalized, but that was not the case, so I am sorry for the confusing signals.

heyleke:
no, its not critical, I'm using a workaround, and indeed, I know how to build a version of the plugin. I was merely wondering if this is normal that this gets stuck for a month...

macetw:
Let's git er done, eh. @heyleke @dwnusbaum

yizhuan:
@heyleke Agree with you. The process is very slow. The first implementation attempt #84 was submitted in Sept 2019. 
Building a self version of the plugin and managing the plugin upgrade is a nightmare. @dwnusbaum

car-roll (approved):

timja:
@car-roll use the checks tab for re-triggering 😄

car-roll:
thanks @timja , old habits die hard 😢

fmuscolino:
I know this is not critical, but are there any updates? I mean, is this planned somehow (so hoping to have a release, let's say, in one month) or it's better to go for a workaround since this for sure won't be considered in next months?  
Just to have an idea of how much effort to put in a workaround
Thanks!

brunzefb:
@heyleke I actually also need this fix.  Can someone point me in the right direction on how to rebuild & install the plugin based on the PR?  It's kind of sad that such a useful feature kind of hangs there with it being 95% done, but I get why this happens. I probably can figure  out how to build, but don't have a lot of experience building java stuff, so it would take longer - just need top level hints.  What is the name of the plugin this goes to, Pipeline: SCM step?

car-roll:
my bad, everyone. I forgot about this PR after re-running the tests. I've re-run it a few times and looks like we're getting consistent failure on `org.jenkinsci.plugins.workflow.libs.SCMRetrieverTest.selfTestLibraries`.

brunzefb:
@car-roll So I got the PR checked out and built this in intellij on my mac, but the project does not output an .hpi file that I could install.  Are there some instructions on how incorporate this into a jenkins instance?  Or do I need to build on Windows/Linux?

brunzefb:
> @car-roll So I got the PR checked out and built this in intellij on my mac, but the project does not output an .hpi file that I could install. Are there some instructions on how incorporate this into a jenkins instance? Or do I need to build on Windows/Linux?

Build generates ./target/test-classes/test-dependencies/workflow-cps.hpi -- is that what one installs?

car-roll:
@brunzefb you should see a `workflow-cps-global-lib.hpi` in the `target` folder after a successful build.

jachstet-sea:
I'd like to repeat what @pvasilevich mentioned: Could we please get a short one-line example on how to use the new `libraryPath` parameter when dynamically loading a pipeline library? I'm a bit confused if I have to specify the retriever or if will magically work without since the actual repository as already been checked out before? Thanks!

dwnusbaum:
You can use the Pipeline snippet generator (`https://<your jenkins instance>/pipeline-syntax/`) to figure out how to use steps. Something like `library(identifier: 'foo@master', retriever: legacySCM(scm: scm, libraryPath: 'your/path/here'))` should work. If the `identifier` corresponds to an existing library that is statically configured (e.g. globally or in a folder), then if you omit `retriever`, the configured retriever settings will be used.

PR #102, opened 2020-11-02
Store data on Global Shared Library usage

j3p0uk:
When a global shared library function is invoked add to an action so that
the data can be found at a later date.


j3p0uk:
@dwnusbaum looking to enhance the reporting from the global shared libs.  Built/tested/works, and would love to get it approved, or take recommendations for what to do in order to get it merged, thanks!

j3p0uk:
https://issues.jenkins.io/browse/JENKINS-64301

src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariable.java
```diff
@@ -66,6 +92,16 @@ public Object getValue(@Nonnull CpsScript script) throws Exception {
                we would also need to start calling LoadStepExecution.Replacer.
             */
             binding.setVariable(getName(), instance);
+
+            if (this.run != null) {
```
dwnusbaum, line 96: Instead of threading the `Run` through the constructor, you could use `CpsScript.$build()`.

src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariableAction.java
```diff
@@ -0,0 +1,139 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@ExportedBean(defaultVisibility = 999)
+public class UserDefinedGlobalVariableAction implements RunAction2 {
+
+    private transient Run run;
+    public List<UserDefinedGlobalVariable> userGlobalVars = new ArrayList<>();
```
dwnusbaum, line 58: Right now this type cannot be serialized safely because `UserDefinedGlobalVariable.run` is non-transient. You can use my recommendation above to remove `UserDefinedGlobalVariable.run` entirely, but I would also recommend introducing a new inner class here that holds whatever properties are necessary so that future changes to `UserDefinedGlobalVariable` do not need to take serialization compatibility into account.

dwnusbaum (changes requested):
@j3p0uk Thanks for the PR! I currently have very little time to evaluate changes to OSS Jenkins plugins apart from bug fixes, but in general I think this would be a nice change and the design looks good other than a few issues which I noted in review comments.

j3p0uk:
> @j3p0uk Thanks for the PR! I currently have very little time to evaluate changes to OSS Jenkins plugins apart from bug fixes, but in general I think this would be a nice change and the design looks good other than a few issues which I noted in review comments.

All changes requested made, thanks for the pointers, I feel better about the code after that! :-)

Failing test looks unrelated to the changes, as the previously passing revision is also fail.ing in the same way in local testing.  Validated that the change still works in testing against jenkins/jenkins:lts docker image.

j3p0uk:
@dwnusbaum does your change in #107 suggest any changes here?

src/main/java/org/jenkinsci/plugins/workflow/cps/global/UserDefinedGlobalVariableAction.java
```diff
@@ -0,0 +1,160 @@
...
+        this.run = run;
+    }
+
+    public Run getRun() {
+        return run;
+    }
+
+    public Api getApi() { return new Api(this); }
+
+    public static class UserDefinedGlobalVariableData {
```
dwnusbaum, line 145: @j3p0uk I would add `@ExportedBean` to this class and add `@Exported` to its getter methods, and then in the parent class I would delete `apiUserGlobalVars` and add `@Exported` to `getUserGlobalVars` instead.

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b track_shared
git pull https://github.com/j3p0uk/workflow-cps-global-lib-plugin track_shared
```

PR #104, opened 2021-02-16
Fix git snippet test

MarkEWaite:
## Fix the git snippet test for git plugin 4.6.0

The git snippet test will fail when the git plugin dependency is updated to git plugin 4.6.0.  This pull request shows one way to fix the test.

### Require Jenkins 2.222.4 or newer

Recent releases of the git plugin require 2.222.4 as their minimum Jenkins version.  Jenkins 2.235.1 is the version recommended by https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/

### Adapt snippet tests for git plugin 4.6.0

[Git plugin 4.6.0](https://github.com/jenkinsci/git-plugin/releases/tag/git-4.6.0) removed the undocumented submodule configurator experiment so that the pipeline syntax snippet generator no longer includes `doGenerateSubmoduleConfigurations` or `submoduleCfg`.  They are accepted as arguments but any values assigned to them are ignored.

The test could have been adapted to run with both older and newer versions of the git plugin, but it seems simpler and easier to show the current behavior.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,6 +188,7 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>git</artifactId>
             <scope>test</scope>
+            <version>4.6.0</version> <!-- Temporary version number until it is included in the Jenkins bom -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.18</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.4</jenkins.version>
+        <jenkins.version>2.222.4</jenkins.version>
         <java.level>8</java.level>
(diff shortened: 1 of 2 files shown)


jglick (approved):

jetersen (approved):

car-roll (approved):

jglick:
@car-roll are you going to release this?

A user was also asking when #100 would get a release.

car-roll:
I was planning to  once  I got permissions to do  a  release. Forgot  to file the pr to do that :(

jglick:
Hmm? You are in https://github.com/orgs/jenkinsci/teams/workflow-cps-global-lib-plugin-developers/members and https://github.com/jenkins-infra/repository-permissions-updater/blob/9ae149730109bd43d7ec922b2e18de6333e79931/permissions/plugin-workflow-cps-global-lib.yml#L13 so I think you are good to go?

car-roll:
2.18 has been released

PR #106, opened 2021-04-14
[JENKINS-62813] Go back to `f:dropdownDescriptorSelector` after tables-to-divs

jglick:
[JENKINS-62813](https://issues.jenkins.io/browse/JENKINS-62813): more or less reverts #17 to give better behavior now that we no longer need to worry about excessive indentation. Before (in 2.277.x):
before
After:
after


timja (approved):

jtnord (approved):

src/main/resources/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever/config.properties
```diff
@@ -1,4 +1,3 @@
-<div>
-    Loads a library from an SCM plugin using newer interfaces optimized for this purpose.
+blurb=\
```
jglick, line 1: Switching from `help.html` to always-visible `f:description` because current versions of Jenkins do not render `help.html` in a dropdown. https://github.com/jenkinsci/jenkins/pull/5417 fixes that, though the appearance is not great (there is no natural place to display the **?** button so we have to make an empty pseudo-row) and we may prefer the `f:description` anyway.

PR #107, opened 2021-04-29
Expose LibrariesAction and LibraryRecord to Jenkins' REST API

dwnusbaum:
This PR exposes data about shared libraries used by a build through the Jenkins REST API. The information was already being stored in `LibrariesAction` and `LibraryRecord`, it was just not exposed to the API. Somewhat similar to #102, which I also think is a good idea. 

I did not add any tests in this PR that use the API, but I could do so if desired.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
@@ -33,11 +33,14 @@ import hudson.model.TaskListener;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
 
 /**
  * A run action recording libraries used in a given build.
  */
-class LibrariesAction extends InvisibleAction {
+@ExportedBean
+public class LibrariesAction extends InvisibleAction {
 
     private final List<LibraryRecord> libraries;
 
@@ -48,6 +51,7 @@ class LibrariesAction extends InvisibleAction {
     /**
      * A list of libraries in use.
      */
+    @Exported
     public List<LibraryRecord> getLibraries() {
         return Collections.unmodifiableList(libraries);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/LibrariesAction.java
```diff
@@ -33,11 +33,14 @@
...
 import java.util.List;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
 
 /**
  * A run action recording libraries used in a given build.
  */
-class LibrariesAction extends InvisibleAction {
+@ExportedBean
+public class LibrariesAction extends InvisibleAction {
```
dwnusbaum, line 43: This class was not previously public. I could add `@Restricted(NoExternalUse.class)` to prevent direct Java-level access from other plugins if desired, but the class seems pretty harmless and likely to be stable going forward so it seemed ok to me to just make it public.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRecord.java
```diff
@@ -24,13 +24,17 @@
...
 import java.util.TreeSet;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
 
 /**
  * Record of a library being used in a particular build.
  */
-final class LibraryRecord {
+@ExportedBean
+public final class LibraryRecord {
```
dwnusbaum, line 37: ditto

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryRecord.java
```diff
@@ -45,6 +49,31 @@
...
+
+    @Exported
+    public boolean isTrusted() {
+        return trusted;
+    }
+
+    @Exported
+    public boolean isChangelog() {
+        return changelog;
+    }
```
dwnusbaum, line 75: This (and `isTrusted` to some extent) do not seem likely to be particularly interesting to external consumers of the REST API, so I could remove the `@Exported` methods for them if anyone feels strongly about it.

jglick (approved):

car-roll (approved):

PR #110, opened 2021-05-11
[JENKINS-65612] Support for immediately cleaning `@libs` workspaces

basil:
See [JENKINS-65612](https://issues.jenkins.io/browse/JENKINS-65612). `SCMRetriever` and `SCMSourceRetriever` create a `@libs` workspace for retrieving the shared library before copying it into the build directory. While old build directories are eventually discarded based on the global build discarder configuration, these `@libs` workspaces tend to stick around for a long time and eventually need to be removed. It would be nice if a flag could be set on `SCMRetriever` and `SCMSourceRetriever` to clean these `@libs` workspaces immediately after we are done with them.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMRetriever.java
@@ -45,6 +45,7 @@ import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * Uses legacy {@link SCM} to check out sources based on variable interpolation.
@@ -52,6 +53,7 @@ import org.kohsuke.stapler.DataBoundConstructor;
 public class SCMRetriever extends LibraryRetriever {
 
     private final SCM scm;
+    private boolean cleanWorkspace;
 
     @DataBoundConstructor public SCMRetriever(SCM scm) {
         this.scm = scm;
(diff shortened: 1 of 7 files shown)


jglick:
Why not just implement `ItemListener.onDeleted` so `workspace/xxx@libs/` gets deleted when `jobs/xxx/` is deleted?

basil:
> just

I could see an argument for _also_ implementing that _by default_ under the general principle of not leaking resources. But I think this PR has independent value. Let me explain why.

We (unfortunately) use a lot of static agents, and we want to keep builds [as isolated as possible](https://twitter.com/tyvole/status/1245376277667885058). So our shared library has a `vars/nodeWithCleanup.groovy` with

```groovy
def call(Map params = [:], Closure closure) {
  def cleanupClosure = {
    try {
      it()
    } finally {
      cleanWs()
    }
  }

  node(params.label) {
    if (params.workspace) {
      ws(params.workspace) {
        cleanupClosure(closure)
      }
    } else {
      cleanupClosure(closure)
    }
  }
}
```

which we use universally as a generic substitute for `node`. This allows us to reclaim the workspace _immediately_ when the run ends, since we were going to wipe it on the next run anyway. This avoids wasting space while jobs are at rest, which is useful for jobs that run very infrequently (we have thousands of jobs that run a few times a month to cover various portions of our test matrix).

We've found the above quite useful, so I also think we would find the feature in this PR useful for the same reason (opt-in, of course, rather than by default).

basil:
> Why not just implement `ItemListener.onDeleted` so `workspace/xxx@libs/` gets deleted when `jobs/xxx/` is deleted?

Done in #115.

car-roll:
#115 merged, closing this pr

jglick:
@car-roll read https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/110#issuecomment-840737198

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b JENKINS-65612
git pull https://github.com/basil/workflow-cps-global-lib-plugin JENKINS-65612
```

PR #115, opened 2021-09-13
[JENKINS-66629] `SCMSourceRetriever` leaks `@libs` workspaces

basil:
### Problem

See [JENKINS-66629](https://issues.jenkins.io/browse/JENKINS-66629). `SCMSourceRetriever` creates a `@libs` workspace for retrieving the shared library before copying it into the build directory. While old build directories are eventually discarded based on the global build discarder configuration, these `@libs` workspaces are leaked when the job is renamed or deleted.

### Solution

When a job's location changes or is deleted, delete the old (now leaked) `@libs` workspace. If the job's location changed, a new `@libs` workspace will be created at the new location the next time the job runs.

### Implementation

As suggested in [jenkinsci/workflow-cps-global-lib-plugin#110 (comment)](https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/110#issuecomment-840695808), implement `ItemListener#onDeleted` so `workspace/xxx@libs/` gets deleted when `jobs/xxx/` is deleted.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
@@ -326,22 +326,20 @@ public class SCMSourceRetriever extends LibraryRetriever {
                                     Jenkins.get().getRawWorkspaceDir(),
                                     itemFullName,
                                     item.getRootDir().getPath());
-                    FilePath baseWorkspace = new FilePath(new File(base));
-                    if (baseWorkspace != null) {
-                        FilePath dir = baseWorkspace.withSuffix(getFilePathSuffix() + "libs");
-                        try {
-                            if (dir.isDirectory()) {
-                                LOGGER.log(
-                                        Level.INFO,
-                                        () -> "Deleting obsolete shared library workspace " + dir);
-                                dir.deleteRecursive();
-                            }
-                        } catch (IOException | InterruptedException e) {
+                    FilePath dir =
+                            new FilePath(new File(base)).withSuffix(getFilePathSuffix() + "libs");
+                    try {
+                        if (dir.isDirectory()) {
                             LOGGER.log(
-                                    Level.WARNING,
-                                    e,
-                                    () -> "Could not delete obsolete shared library workspace " + dir);
+                                    Level.INFO,
+                                    () -> "Deleting obsolete shared library workspace " + dir);
+                            dir.deleteRecursive();
                         }
+                    } catch (IOException | InterruptedException e) {
+                        LOGGER.log(
+                                Level.WARNING,
+                                e,
+                                () -> "Could not delete obsolete shared library workspace " + dir);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
```diff
@@ -293,4 +301,62 @@ public FormValidation doCheckLibraryPath(@QueryParameter String libraryPath) {
...
+                            .getName()
+                            .equals("org.jenkinsci.plugins.workflow.job.WorkflowJob")) {
+                synchronized (item) {
+                    String base =
+                            expandVariablesForDirectory(
+                                    Jenkins.get().getRawWorkspaceDir(),
+                                    itemFullName,
+                                    item.getRootDir().getPath());
+                    FilePath baseWorkspace = new FilePath(new File(base));
+                    if (baseWorkspace != null) {
```
jglick, line 330: How could it be null if you just constructed it?

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
```diff
@@ -293,4 +301,62 @@ public FormValidation doCheckLibraryPath(@QueryParameter String libraryPath) {
...
+                                    Level.WARNING,
+                                    e,
+                                    () -> "Could not delete obsolete shared library workspace " + dir);
+                        }
+                    }
+                }
+            }
+        }
+
+        private static String expandVariablesForDirectory(
```
jglick, line 350: I guess we need to do this because if the item is moved, it will be too late to look up its original workspace location by the time we get the notification?

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
basil: Silly me. The null check was left over from an older version of this code and I forgot to remove it after refactoring. I'll do that now.

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
basil: Yes, exactly.

jglick: Took me at least a while to get there…maybe deserves a comment, linking to https://github.com/jenkinsci/jenkins/blob/f03183ab09ce5fb8f9f4cc9ccee42a3c3e6b2d3e/core/src/main/java/jenkins/model/Jenkins.java#L2567-L2576? (And maybe we should just remove the API restriction there so that at some point this code could be replaced with a call to the official method?)

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/libs/SCMSourceRetriever.java
basil: I updated the comment. I feel that there are enough bits and pieces of core duplicate in Pipeline that this isn't materially changing the situation. That and the… "overhead" of making API changes to core makes me want to leave this alone.

car-roll (approved):

PR #116, opened 2021-09-20
Allow explicit selection of default library tag.

dvpierce:
If developer manually specifies the default version of a library where overriding the default version is not allowed, this should allow the checkout to proceed normally.

https://issues.jenkins.io/browse/JENKINS-66682

- [] Link to relevant pull requests, esp. upstream and downstream changes
- [] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryConfiguration.java
@@ -140,7 +140,7 @@ public class LibraryConfiguration extends AbstractDescribableImpl<LibraryConfigu
     }
 
     @Nonnull String defaultedVersion(@CheckForNull String version) throws AbortException {
-        if (version == null) {
+        if ( (version == null) || (version.equals(defaultVersion)) ) {
             if (defaultVersion == null) {
                 throw new AbortException("No version specified for library " + name);
             } else {


car-roll:
I'm hesitant to go this route as it can cause more confusion for users that are not aware of this behavior. At the very least there needs to be some logging to explain that this has happened and possibly even a flag to enable this behavior.

car-roll (changes requested):
I'm hesitant to go this route as it can cause more confusion for users that are not aware of this behavior. At the very least there needs to be some logging to explain that this has happened and possibly even a flag to enable this behavior.

dvpierce:
Hello @car-roll - I have added logging as you suggested.

I'm not familiar enough with the codebase to know how I might create a toggle/option for this functionality - could you give me a hint?

Please LMK your thoughts.

car-roll:
@dvpierce Something like this: https://github.com/jenkinsci/durable-task-plugin/blob/adeab44bbb208cdd3e375775eacd281e6e49797b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java#L69

You would then be able to set the variable in the jenkins command-line startup or via the script console.

car-roll (approved):

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b patch-1
git pull https://github.com/dvpierce/workflow-cps-global-lib-plugin patch-1
```

PR #119, opened 2021-11-04
Update parent pom and core requirement

alecharp:
Kind of a follow up of jenkinsci/jenkins#5320.

ping @car-roll 

- [x Ensure that the pull request title represents the desired changelog entry


pom.xml
```diff
@@ -64,16 +64,16 @@
     <properties>
         <revision>2.22</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.277.1</jenkins.version>
+        <jenkins.version>2.303.1</jenkins.version>
```
car-roll, line 67: 2.303 does seem like a big bump, but it does seem like the majority of users are on that.

car-roll (approved):

bitwiseman (approved):
The 2.303.x line is the current LTS line, so I'd generally rather not force people to it, but there aren't other changes to be released. This seems fine to me.

car-roll:
Yikes, I accidentally bumped jenkins to 2.304 in #120, fixing here and reflecting change in logs

pom.xml
jglick: FTR: https://github.com/jenkinsci/jenkins/pull/5320 is first in 2.303.x as an LTS line, so this is not arbitrary.

jglick:
> Kind of a follow up of https://github.com/jenkinsci/jenkins/pull/5320.

@alecharp @car-roll can you explain in what sense this is a follow-up to that core change? This baseline is causing me some problems in handling #172 so I would like to know if it was truly necessary or if I can in fact revert the baseline to, say, 2.289.x at least for one release.

PR #122, opened 2021-11-10
[JENKINS-60406] Replace RUN_SCRIPTS with ADMINISTER

bitwiseman:
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant pull requests, esp. upstream and downstream changes


car-roll (approved):

workflow-cps-global-lib.iml
```diff
@@ -0,0 +1,249 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jglick, line 1: Can you ignore this please?

jglick (changes requested):
Forgot https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/a1de00ea889a91e3cb4d143c0ba5efb6a7815db3/src/main/resources/org/jenkinsci/plugins/workflow/libs/GlobalLibraries/config.jelly#L28

jtnord (approved):
Looks good to me.

It may be the code can be further cleaned up as global configuration should never be called unless you have administer, but that will open a can of worms so this looks like a reasonable step that can be potentially refined at some point in the future.

olamy (approved):
LGTM (no much more to say than @jglick comments)

bitwiseman:
@jglick Addressed feedback.

car-roll:
looks like @jglick 's comment has been addressed, going ahead to merge

PR #139, opened 2022-02-16
Fix recent security fixes if JENKINS_HOME is a symlink and update tests for compatibility with Windows and Git plugin 4.10.3

dwnusbaum:
The tests for security fixes for https://www.jenkins.io/security/advisory/2022-02-15/ fail against recent version of Git plugin due to https://github.com/jenkinsci/git-plugin/pull/1207, and also fail when running on Windows due to a difference in the behavior.

The fix itself also did not allow `JENKINS_HOME` to be a symlink, because it compared a canonical path to an absolute path.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -256,7 +256,7 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
                 for (LibraryRecord library : action.getLibraries()) {
                     FilePath libResources = libs.child(library.getDirectoryName() + "/resources/");
                     FilePath f = libResources.child(name);
-                    if (!new File(f.getRemote()).getCanonicalFile().toPath().startsWith(libResources.absolutize().getRemote())) {
+                    if (!new File(f.getRemote()).getCanonicalFile().toPath().startsWith(new File(libResources.getRemote()).getCanonicalPath())) {
                         throw new AbortException(name + " references a file that is not contained within the library: " + library.name);
                     } else if (f.exists()) {
                         resources.put(library.name, readResource(f, encoding));
--- a/src/test/java/org/jenkinsci/plugins/workflow/libs/ResourceStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/libs/ResourceStepTest.java
@@ -189,6 +189,7 @@ public class ResourceStepTest {
 
     @Issue("SECURITY-2479")
     @Test public void symlinksInLibraryResourcesAreNotAllowedToEscapeWorkspaceContext() throws Exception {
+        assumeFalse(Functions.isWindows()); // On Windows, the symlink is treated as a regular file, so there is no vulnerability, but the behavior is different.
         sampleRepo.init();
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -256,7 +256,7 @@
                 for (LibraryRecord library : action.getLibraries()) {
                     FilePath libResources = libs.child(library.getDirectoryName() + "/resources/");
                     FilePath f = libResources.child(name);
-                    if (!new File(f.getRemote()).getCanonicalFile().toPath().startsWith(libResources.absolutize().getRemote())) {
+                    if (!new File(f.getRemote()).getCanonicalFile().toPath().startsWith(new File(libResources.getRemote()).getCanonicalPath())) {
```
dwnusbaum, line 259: The only intended difference here is that if `JENKINS_HOME` is a symlink, the code should no longer throw an exception in non-dangerous scenarios.

jglick, line 259:
```suggestion
                    if (!new File(f.getRemote()).getCanonicalFile().toPath().startsWith(new File(libResources.getRemote()).getCanonicalFile().toPath())) {
```
?

jglick: Though even better is to use NIO fully:
```suggestion
                    if (!Paths.get(f.getRemote()).toRealPath().startsWith(Paths.get(libResources.getRemote()).toRealPath())) {
```

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
dwnusbaum: `Path.toRealPath` throws if the path does not exist on the file system, so we would have to change things a bit more if we wanted to keep attackers from being able to probe the existence of files using symlinks.

Is there a meaningful difference between `Path.startsWith(String)` and `Path.startsWith(Path)`? They both return false for things like `foo/bar` starting with `f`.

dwnusbaum: > Mm, I see. So there is no NIO equivalent to File.getCanonicalPath?

Not directly as far as I know. I did consider using NIO in the original patch and catching `NoSuchFileException` so that error messages would be the same regardless of file existence, but it seemed simpler to use `getCanonicalFile`.

> I do not know.

Ok, I just wasn't sure if it was a style-related suggestion or if I was missing an important detail. As far as I know the current code with `Path.startsWith(f.getCanonicalPath())` and the proposed code with `Path.startsWith(f.getCanonicalFile().toPath())` should have the same behavior.

jglick: Probably. The weird corner cases are likely to be on Windows. I think the current code is secure and likely to fix the regression for most if not all users.

dwnusbaum: Ok, thanks for elaborating. I'll merge the code as-is and will follow up as needed.

PR #147, opened 2022-03-15
convert library cache exclude to substring search

JeffCobble:
Converting the library cache code to be a substring search vice explicit string.  This enables blanket exemptions for things like feature branches without breaking current functionality. This is required to enable caching for branches like master and exclude others without having to know the full name as a feature branch would be dynamic.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

All tests run and caching works as expected with change.  Unfortunately I cannot upload screenshots due to work requirements. I validated all build tests worked as well.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration.java
@@ -53,15 +53,14 @@ public final class LibraryCachingConfiguration extends AbstractDescribableImpl<L
     }
 
     public Boolean isExcluded(String version) {
-        return getExcludedVersions().contains(version);
-        // try{
-        //     for (String it : getExcludedVersions()){
-        //         if ((version.contains(it)) && (it != "")){
-        //             return true;
-        //         }
-        //     }
-        // }finally{}
-        // return false;
+        if(version != null){
+            for (String it : getExcludedVersions()){
+                if ((version.contains(it)) && (it != "")){
+                    return true;
+                }
+            }
+        }
+        return false;
     }
 
     @Override public String toString() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration.java
@@ -53,7 +53,12 @@ public final class LibraryCachingConfiguration extends AbstractDescribableImpl<L
     }
 
     public Boolean isExcluded(String version) {
-        return getExcludedVersions().contains(version);
+        for (String it : getExcludedVersions()){
+            if (version.contains(it)){
+                return true;
+            }
+        }
+        return false;
     }
 
     @Override public String toString() {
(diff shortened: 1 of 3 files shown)


JeffCobble:
@dwnusbaum Do you mind taking a look at this?  I believe the failing tests are not local to this PR as another PR has the exact same failures.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration.java
```diff
@@ -53,7 +53,12 @@ public String getExcludedVersionsStr() {
     }
 
     public Boolean isExcluded(String version) {
-        return getExcludedVersions().contains(version);
+        for (String it : getExcludedVersions()){
+            if (version.contains(it)){
```
dwnusbaum, line 57: This line is why `LibraryCachingConfigurationTest.isExcluded` is failing with an NPE: `version` may be null.

It's also why `ResourceStepTest.caching` is failing: when `excludedVersionsStr` is `""`, `getExcludedVersions()` returns a set with a single `""`, and `version.contains("")` is true for every non-null version.

dwnusbaum (changes requested):
As far as I can tell, the test failures are caused by your changes.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration.java
JeffCobble: @dwnusbaum dwnusbaum, PR should be ready for review

JeffCobble:
@jglick Would you mind giving this PR a review if you have time?

jglick:
I am not maintaining this.

car-roll:
@JeffCobble LGTM, but could you add an additional test to https://github.com/jenkinsci/workflow-cps-global-lib-plugin/blob/e62a4ebbe039329342b51319eb43dbc7d5edaf11/src/test/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfigurationTest.java that covers this condition?

Basically just another `MULTIPLE_EXCLUDED_VERSIONS` that would succeed only with this PR.

JeffCobble:
> 
@car-roll test cases added as requested.  Please ignore the failures as unfortunately I'm having issues running locally so had to use a push to confirm they worked as they should.

car-roll (approved):
I think @dwnusbaum 's comment has been addressed, not sure if he has anything to add to this.
Looks fine to be, but I don't know if I am missing any nuance here with these changes. Thoughts?

If anyone doesn't say anything, I'll merge this next week.

JeffCobble:
@car-roll are we good to merge this one?  Not seeing any comments by anyone.

daniel-beck:
How would one exclude `master` but not `not-master`?

PR #148, opened 2022-03-18
Fix library cache deletion after SECURITY-2586

dwnusbaum:
The security fixes related to the cache feature in https://github.com/jenkinsci/workflow-cps-global-lib-plugin/commit/ace0de3c2d691662021ea10306eeb407da6b6365 broke `LibraryCachingCleanup` (cleans up unused cache directories) and `LibraryCachingConfiguration.doClearCache` (allows admins to manually delete all cached versions of a specific library). This went unnoticed because neither piece of functionality had any test coverage.

This PR fixes the issues with library cache deletion after SECURITY-2586 and adds relevant test coverage.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
@@ -190,7 +190,6 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
 
             if(versionCacheDir.exists()) {
                 listener.getLogger().println("Library " + name + "@" + version + " is cached. Copying from home.");
-                lastReadFile.touch(System.currentTimeMillis());
             } else {
                 listener.getLogger().println("Caching library " + name + "@" + version);
                 versionCacheDir.mkdirs();
@@ -198,6 +197,8 @@ import org.jenkinsci.plugins.workflow.flow.FlowCopier;
                 retriever.retrieve(name, version, changelog, versionCacheDir, run, listener);
                 retrieveLockFile.delete();
             }
+            lastReadFile.touch(System.currentTimeMillis());
+            versionCacheDir.withSuffix("-name.txt").write(name, "UTF-8");
             versionCacheDir.copyRecursiveTo(libDir);
         } else {
             retriever.retrieve(name, version, changelog, libDir, run, listener);
--- a/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingCleanup.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingCleanup.java
@@ -8,29 +8,47 @@ import hudson.model.TaskListener;
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 
+import jenkins.util.SystemProperties;
+
 @Extension public class LibraryCachingCleanup extends AsyncPeriodicWork {
-    private final Long recurrencePeriod;
-    private final Long unreadCacheClearTime;
+    public static /* non-final for script console */ int EXPIRE_AFTER_READ_DAYS =
+            SystemProperties.getInteger(LibraryCachingCleanup.class.getName() + ".CACHE_EXPIRY_DAYS", 7);
 
     public LibraryCachingCleanup() {
         super("LibraryCachingCleanup");
-        recurrencePeriod = Long.getLong("jenkins.workflow-libs.cacheCleanupPeriodDays", TimeUnit.DAYS.toMillis(7));
-        unreadCacheClearTime = Long.getLong("jenkins.workflow-libs.unreadCacheClearDays", TimeUnit.DAYS.toMillis(7));
     }
 
     @Override public long getRecurrencePeriod() {
-        return recurrencePeriod;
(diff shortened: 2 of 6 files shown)


src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -190,14 +190,15 @@
...
                 listener.getLogger().println("Library " + name + "@" + version + " is cached. Copying from home.");
-                lastReadFile.touch(System.currentTimeMillis());
             } else {
                 listener.getLogger().println("Caching library " + name + "@" + version);
                 versionCacheDir.mkdirs();
                 retrieveLockFile.touch(System.currentTimeMillis());
                 retriever.retrieve(name, version, changelog, versionCacheDir, run, listener);
                 retrieveLockFile.delete();
             }
+            lastReadFile.touch(System.currentTimeMillis());
```
dwnusbaum, line 200: Now we write this even when creating the cache directory for the first time, so that it will eventually be deleted even if it was only used once.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -190,14 +190,15 @@
...
+            versionCacheDir.withSuffix("-name.txt").write(name, "UTF-8");
```
dwnusbaum, line 201: We write this every time rather than just when the cache is first created to handle caches created after the security fix but before this PR.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingCleanup.java
```diff
@@ -8,29 +8,47 @@
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 
+import jenkins.util.SystemProperties;
+
 @Extension public class LibraryCachingCleanup extends AsyncPeriodicWork {
-    private final Long recurrencePeriod;
-    private final Long unreadCacheClearTime;
+    public static /* non-final for script console */ int EXPIRE_AFTER_READ_DAYS =
+            SystemProperties.getInteger(LibraryCachingCleanup.class.getName() + ".CACHE_EXPIRY_DAYS", 7);
```
dwnusbaum, line 15:
```suggestion
            SystemProperties.getInteger(LibraryCachingCleanup.class.getName() + ".EXPIRE_AFTER_READ_DAYS", 7);
```

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingCleanup.java
```diff
@@ -8,29 +8,47 @@
...
 
     public LibraryCachingCleanup() {
         super("LibraryCachingCleanup");
-        recurrencePeriod = Long.getLong("jenkins.workflow-libs.cacheCleanupPeriodDays", TimeUnit.DAYS.toMillis(7));
-        unreadCacheClearTime = Long.getLong("jenkins.workflow-libs.unreadCacheClearDays", TimeUnit.DAYS.toMillis(7));
```
dwnusbaum, line 18: These did not follow standard Jenkins system property naming conventions, and I did not see any documentation related to them, so I created a new `EXPIRE_AFTER_READ_DAYS` property that follows the normal Jenkins style.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingCleanup.java
```diff
@@ -8,29 +8,47 @@
...
     }
 
     @Override public long getRecurrencePeriod() {
-        return recurrencePeriod;
+        return TimeUnit.HOURS.toMillis(12);
```
dwnusbaum, line 22: Previously this was 7 days, which combined with `unreadCacheClearTime` being 7 days meant that in the worst case a library could go almost 2 weeks without being cleaned up, and if Jenkins restarts in that time frame, the timer would start all over again.

Since the check is not that expensive, I think it is fine to run it twice a day.

src/main/resources/org/jenkinsci/plugins/workflow/libs/LibraryCachingConfiguration/config.jelly
```diff
@@ -31,5 +31,7 @@ THE SOFTWARE.
     <f:entry title="${%Versions to exclude}" field="excludedVersionsStr">
         <f:textbox />
     </f:entry>
-    <f:validateButton title="${%Clear cache}" progress="${%Clearing...}" method="clearCache" with="name" />
+    <j:if test="${h.hasPermission(app.ADMINISTER)}">
+        <f:validateButton title="${%Clear cache}" progress="${%Clearing...}" method="clearCache" with="name" />
+    </j:if>
```
dwnusbaum, line 36: The backend has always required `ADMINISTER` permission, but we used to show this button on folder-level libraries to anyone with Job/Configure permission, which is just confusing.

car-roll (approved):

mawinter69:
Any idea when this will get merged?
I need this for my fix in #151

jtnord (approved):

jtnord (approved):

PR #151, opened 2022-03-30
[fix JENKINS-66898] make the cache thread safe

mawinter69:
Fix [JENKINS-66898](https://issues.jenkins.io/browse/JENKINS-66898)

there are several conditions where we have race conditions in
the shared library cache :
- when multiple builds start at the same time and try to use an expired
  cached shared library. The first one will start
  deleting the cache dir. The second one will see that the cachedir was
  modified and consider it up-to-date while the first is still deleting
- when multiple builds start at the same time with no cache entry available
  The first has created the cachedir but not yet the lock file, the second will not see the lockfile
  but the cachedir
- the background cleaner is about to clean when a new build starts
- an administrator decides to clear the cache right before a new build
  wants to use the cache library

This PR fixes the race conditions by using a ReadWriteLock to ensure nobody reads when a cache entry is created/updated/deleted

Needs a rework after #148 is merged

testing race conditions is extremely hard so no tests.


src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryCachingCleanup.java
```diff
@@ -49,8 +50,15 @@ private boolean removeIfExpiredCacheDirectory(FilePath library) throws IOExcepti
         final FilePath lastReadFile = new FilePath(library, LibraryCachingConfiguration.LAST_READ_FILE);
         if (lastReadFile.exists()) {
             if (System.currentTimeMillis() - lastReadFile.lastModified() > TimeUnit.DAYS.toMillis(EXPIRE_AFTER_READ_DAYS)) {
-                library.deleteRecursive();
-                library.withSuffix("-name.txt").delete();
+        	ReentrantReadWriteLock retrieveLock = LibraryAdder.getReadWriteLockFor(library.getName());
+                if (retrieveLock.writeLock().tryLock()) {
```
dwnusbaum, line 54: I think it is fine to block interruptibly waiting for the lock here:

```suggestion
                if (retrieveLock.writeLock().lockInterruptibly()) {
```

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -164,42 +207,61 @@ static List<URL> retrieve(@NonNull LibraryRecord record, @NonNull LibraryRetriev
...
+        
         if(shouldCache) {
-            if (cachingConfiguration.isRefreshEnabled()) {
-                final long cachingMinutes = cachingConfiguration.getRefreshTimeMinutes();
-                final long cachingMilliseconds = cachingConfiguration.getRefreshTimeMilliseconds();
-
-                if(versionCacheDir.exists() && (versionCacheDir.lastModified() + cachingMilliseconds) < System.currentTimeMillis()) {
-                    listener.getLogger().println("Library " + name + "@" + version + " is due for a refresh after " + cachingMinutes + " minutes, clearing.");
-                    versionCacheDir.deleteRecursive();
+            retrieveLock.readLock().lock();
```
dwnusbaum, line 221: I think we should use `lockInterruptibly` if we are going to block while acquiring these locks.

```suggestion
            retrieveLock.readLock().lockInterruptibly();
```

dwnusbaum (commented):
Hi, thanks for the PR! I took a quick look and added some comments.

src/main/java/org/jenkinsci/plugins/workflow/libs/LibraryAdder.java
```diff
@@ -64,7 +64,19 @@
 @Extension public class LibraryAdder extends ClasspathAdder {
 
     private static final Logger LOGGER = Logger.getLogger(LibraryAdder.class.getName());
+    
+    private static Map<String, ReentrantReadWriteLock> cacheRetrieveLock = new HashMap<>();
```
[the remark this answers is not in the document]
mawinter69: I disagree. A (global) `ReentrantLock` would serialize all library retrieval operations, that could lead to noticeable longer delays when many jobs start at the same time.
And a global `ReentrantReadWriteLock` would make only a marginal difference in the code.
And I had the impression performance is important looking at lines 175 to 178 of LibraryAdder (before my changes)

jglick:

```bash
cd pipeline-groovy-lib-plugin
git checkout -b JENKINS-66898
git pull https://github.com/mawinter69/workflow-cps-global-lib-plugin JENKINS-66898
```

mawinter69:
Opened PR https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/3 in the new plugin

PR #169, opened 2022-05-18
Fix test broken by signature changes related to SECURITY-2450

yaroslavafenkin:
`CpsFlowDefinition#doCheckScript`'s signature was changed in https://github.com/jenkinsci/workflow-cps-plugin/commit/6bd4e8bf837ba1df4394efdd003c70bb1b9ac33f.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryLocalTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryLocalTest.java
@@ -65,7 +65,7 @@ public class WorkflowLibRepositoryLocalTest extends Assert {
         assertEquals(cfdd.doCheckScriptCompile(p, "import org.acme.Foo"), CpsFlowDefinitionValidator.CheckStatus.SUCCESS.asJSON());
         assertNotEquals(cfdd.doCheckScriptCompile(p, "import org.acme.NoSuchThing").toString(), CpsFlowDefinitionValidator.CheckStatus.SUCCESS.asJSON().toString()); // control test
         // valid from script-security point of view
-        assertSame(cfdd.doCheckScript("import org.acme.Foo", true), FormValidation.ok());
-        assertSame(cfdd.doCheckScript("import org.acme.NoSuchThing", true), FormValidation.ok());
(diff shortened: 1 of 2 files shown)


dwnusbaum (approved):

src/test/java/org/jenkinsci/plugins/workflow/cps/global/WorkflowLibRepositoryLocalTest.java
```diff
@@ -65,7 +65,7 @@ public void initialRepoShouldBeEmpty() throws Exception {
         assertEquals(cfdd.doCheckScriptCompile(p, "import org.acme.Foo"), CpsFlowDefinitionValidator.CheckStatus.SUCCESS.asJSON());
         assertNotEquals(cfdd.doCheckScriptCompile(p, "import org.acme.NoSuchThing").toString(), CpsFlowDefinitionValidator.CheckStatus.SUCCESS.asJSON().toString()); // control test
         // valid from script-security point of view
-        assertSame(cfdd.doCheckScript("import org.acme.Foo", true), FormValidation.ok());
-        assertSame(cfdd.doCheckScript("import org.acme.NoSuchThing", true), FormValidation.ok());
+        assertSame(cfdd.doCheckScript("import org.acme.Foo", null, true), FormValidation.ok());
+        assertSame(cfdd.doCheckScript("import org.acme.NoSuchThing", null, true), FormValidation.ok());
```
dwnusbaum, line 69: Looking at this again, I think these assertions are pointless these days since `doCheckScript` is essentially a no-op when `sandbox` as `true`, see https://github.com/jenkinsci/workflow-cps-plugin/blob/8732e3150b97bd4213401814ba72925ebfa86e0d/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsFlowDefinition.java#L163. We could just delete these 3 lines instead and revert the POM changes for a more limited change.

dwnusbaum: See #171.

dwnusbaum:
Fixed by https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/171 instead.

PR #211, opened 2023-05-21
Refresh plugin

basil:


basil:
Used my admin powers to replay the build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Plugins/job/workflow-cps-global-lib-plugin/job/PR-211/2/

PR #213, opened 2024-01-18
Forward compatibility with Guice 7

basil:
Without dropping compatibility with Guice 6 (currently delivered by Jenkins core), add support for Guice 7.


jglick (approved):
Already deprecated so I do not plan to bother replacing `@Inject` with `ExtensionList.lookupSingleton`.

